PR #4, opened 2017-09-13
feat: add the french translation

lynchmaniac:
This is just a little translation in french for the github page for pnpm. I am not necessarily good at translations and they are probably not perfect but it is a working basis. Tell me if you see any incorrect things in this pull request


zkochan:
Awesome!

PR #8, opened 2018-03-14
docs: hooks tips

vjpr:


vjpr:
@zkochan Do I need to `npm run build`?

vjpr:
Local tarball worflow:

```
hub clone facebook/metro
cd metro
lerna bootstrap
mkdir .dev
echo '.dev/' >> .git/info/exclude
cd packages/metro-resolver
npm run prepare-release
cd ../../.dev
npm pack ../packages/metro-resolver
```

zkochan:
The https://github.com/pnpm/pnpm/issues/1054 is only about the root depth. Is metro the metro-resolver a root dep?

Though the update command with depth should update it with forcing, so might be an issue.

vjpr:
@zkochan Can you comment on https://github.com/pnpm/pnpm.github.io/pull/8#issuecomment-373010735

zkochan:
I'd rather not introduce this additional complexity with links in subdeps because it can be solved differently.

You can link the package to the root and use the hook to remove metro-resolver from the dependencies of metro. In that case metro will use metro-resolver which you linked in the root

vjpr:
> You can link the package to the root 

Thanks. I will try it out.

vjpr:
@zkochan 

> You can link the package to the root and use the hook to remove metro-resolver

One caveat is that this doesn't work if there are multiple versions of the package used by different subdeps.

But linking subdeps still could cause issues I guess due to node resolving up from realpaths.

Would it be possible to copy the files in? This would save me a step of having to `npm pack` which is a pretty clunky workflow.

zkochan:
I think maybe it would be fine to allow linking to subdeps but only by absolute path

vjpr:
Perfect!

PR #11, opened 2018-04-16
Update "Who's Using This?" logos

pgonzal:
- Replaced the Rush logo with an SVG version 
- Added a Microsoft logo


pgonzal:
@zkochan if you like we could add logos for each product (SharePoint, Office Fabric, etc).  However it might look weird if we start filling up your list with Microsoft properties.

zkochan:
I've made the logo wider. Looks nicer this way https://github.com/pnpm/pnpm.github.io/commit/60573a1b5d6c0b239241cfaeda10cd3f72f4bccd

PR #17, opened 2018-08-13
DOCS: How to quickly resolve missing deps

o-alexandrov:
## Commits

- ADDED: easiest solution for missing deps

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -53,6 +53,21 @@ dependency should be fixed. That might take time though, so pnpm supports workar
 
 ### Solution 1
 
+In the following example, a dependency does **not** have the `iterall` module in its own list of deps.
+
+The easiest solution to resolve missing dependencies of the buggy packages is to **add `optionalDependencies` to our project's `package.json`**.
+
+```json
+  "devDependencies": {
+    ...
+  },
+  "optionalDependencies": {
+    "iterall": "^1.2.2"
+  }
+```
+
+### Solution 2
+
 One of the solutions is to use [hooks](#hooks) for adding the missing dependencies to the package's `package.json`.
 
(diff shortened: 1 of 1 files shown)


zkochan:
why to optional dependencies, not prod dependencies? Both will work

o-alexandrov:
Thank you for a fast reply. That's cause setting it in `optionalDependencies` you can ensure that you don't mess up other devs' set ups.

Some people tend to use the `--production` flag when they run `npm i --production`.
IMHO, our use case might affect their workflow, how do you think?

o-alexandrov:
@zkochan 
I would suggest not to do so, because there are cases that a missing dependency is a dependency of a devDependency.
Therefore, you would be installing an unnecessary dependency in a production environment.

Please let me know your final decision and I will add what you would like.

zkochan:
I see, but optionalDependencies will be installed in prod as well, so it doesn't really make a difference

o-alexandrov:
You would be able to avoid them in prod by running:
`npm i -production --no-optional`

zkochan:
but the `--no-optional` will also skip optional deps of subdependencies. So I think a better solution would be to save the new dependency with the same type as the type of the dependency that misses it. So if `iterall` is required by a dev dep, install it as a dev dep. If it is required by an optional dep, install it as an optional dep, etc.

o-alexandrov:
Before starting to use `pnpm` today... 😆 

I used to have an alias:
`ni = npm i --no-optional`

You can safely ignore them as they are unrelated to a working state of a module.

zkochan:
Optional dependencies may improve things on some OS. I don't think we have to suggest developers to run pnpm with `--no-optional` because it looks like a limitation of pnpm

o-alexandrov:
Okay, please edit the changes how you see they would fit better.
I agree that you know better this community and I would happily follow your decision :)

PR #20, opened 2018-10-17
docs: add SheetJS to users

SheetJSDev:
in consideration of the fact that BerryNodes contributes to the opencollective https://opencollective.com/pnpm, their logo was moved to the front


zkochan:
ok for now but in long term I'd add a backers section above users.

PR #35, opened 2019-02-19
Semaphore 2.0 CI example configuration

markoa:
Thanks to @qm3ster for [motivating me](https://dev.to/qm3ster/comment/8p06) to write this. :)

### DIFF
--- a/docs/continuous-integration.md
+++ b/docs/continuous-integration.md
@@ -31,7 +31,7 @@ blocks:
   - name: Install dependencies
     task:
       jobs:
-        - name: npm install
+        - name: pnpm install
           commands:
             - curl -L https://unpkg.com/@pnpm/self-installer | node
             - checkout
--- a/docs/continuous-integration.md
+++ b/docs/continuous-integration.md
@@ -16,6 +16,30 @@ install:
   - pnpm install
 ```
 
+## Semaphore
+
+On [Semaphore](https://semaphoreci.com), you can use pnpm for installing and caching your dependencies by adding this to your `.semaphore/semaphore.yml` file:
+
+```yaml
+version: v1.0
+name: Semaphore CI pnpm example
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+blocks:
+  - name: Install dependencies
+    task:
+      jobs:
+        - name: npm install
+          commands:
+            - curl -L https://unpkg.com/@pnpm/self-installer | node
+            - checkout
+            - cache restore node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum package-lock.json),node-modules-$SEMAPHORE_GIT_BRANCH,node-modules-master
+            - pnpm install
+            - cache store node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum package-lock.json) node_modules
+```
+
 ## AppVeyor
 
 On [AppVeyor](https://www.appveyor.com/), you can use pnpm for installing your dependencies by adding this to your `appveyor.yml`:


docs/continuous-integration.md
```diff
@@ -16,6 +16,30 @@ install:
...
+name: Semaphore CI pnpm example
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+blocks:
+  - name: Install dependencies
+    task:
+      jobs:
+        - name: npm install
```
zkochan: probably should be `pnpm install`

markoa: Oh yes, fixed. 🙏

qm3ster:
Oh no, don't go putting me on blast like that :pizza:

qm3ster:
@zkochan shouldn't `checksum package-lock.json` be `checksum shrinkwrap.yaml`?
Additionally, aren't all the files in the store?

I came up with something like this (for a "monorepo" case with `shared-workspace-shrinkwrap = true`):

```yml
version: v1.0
name: JASON DERULO
agent:
  machine:
    type: e1-standard-4
    os_image: ubuntu1804
blocks:
  - name: Install
    task:
      jobs:
        - name: pnpm install
          commands:
            - curl -L https://unpkg.com/@pnpm/self-installer | node
            - checkout
            - cache restore pnpm-store-$SEMAPHORE_GIT_BRANCH-$(checksum shrinkwrap.yaml),pnpm-store-$SEMAPHORE_GIT_BRANCH,pnpm-store-master
            - cache restore node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum shrinkwrap.yaml),node-modules-$SEMAPHORE_GIT_BRANCH,node-modules-master
            - pnpm i
            - pnpm store prune
            - cache store node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum shrinkwrap.yaml) node_modules
            - cache store pnpm-store-$SEMAPHORE_GIT_BRANCH-$(checksum shrinkwrap.yaml) ~/.pnpm-store
  - name: Tests
    task:
      prologue:
        commands:
          - curl -L https://unpkg.com/@pnpm/self-installer | node
          - checkout
          - cache restore pnpm-store-$SEMAPHORE_GIT_BRANCH-$(checksum shrinkwrap.yaml),pnpm-store-$SEMAPHORE_GIT_BRANCH,pnpm-store-master
          - cache restore node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum shrinkwrap.yaml),node-modules-$SEMAPHORE_GIT_BRANCH,node-modules-master
          - pnpm i
      jobs:
        - name: Test
          commands:
            - pnpm m t
```
Pls tell me how dumb I am and how I should improve this.

zkochan:
I don't think you need to cache node_modules. Enough to cache the store. But be aware, sometimes CI caching is slower then just running `pnpm install` (noticed on travis CI)

yes, shrinkwrap.yaml instead of package-lock. But better `pnpm-lock.yaml`.
in pnpm v3 shrinkwrap.yaml is renamed to pnpm-lock.yaml

qm3ster:
Excuse me for not having seen your release from 20 hours ago :v
(Just saw it before this reply, was coming to correct it)
Congratulations and great work, btw :tada: 

I was considering caching the repository's `node_modules` for native module builds, since afaik those at the moment aren't kept in the store?

zkochan:
Thanks!

The builds are in the store if you set the `side-effects-cache` config to `true`
https://pnpm.js.org/docs/en/pnpm-install.html#side-effects-cache

qm3ster:
I saw, but I got scared off by the `stability: Experimental`. With your blessing, I will try it out.

...And here I thought I would delete my `.npmrc` since the only two things in there are now defaults :smile:

zkochan:
Yes, I believe there may be some edge cases but Glitch uses that option, so probably it works well enough

qm3ster:
I *should* be able to do a subsequent install with `--offline` on the same lockfile, right?
It works locally if I just delete `node_modules`, but on the next semaphore step it isn't happy.
It [says the tarballs aren't there](https://qm3ster.semaphoreci.com/jobs/9988672a-c625-45ac-a879-b07fa14eaa0f#L772) (I went into "Debug...", they are there.)

But why is it even looking at the tarballs? Shouldn't it be looking at the unpacked module in the store?

zkochan:
yes, subsequent install with --offline on the same lockfile should work

qm3ster:
Any idea what could go wrong for it to give that error?
Like I said, after the `cache restore pnpm-store-$SEMAPHORE_GIT_BRANCH-$(checksum pnpm-lock.yaml)` line, I manually inspected the store and everything was there.
Is there some kind of env variable/manifest file somewhere that needs to be preserved besides the store directory?

I see the log link might require github login, and not sure if you can see it even then, despite public repo, so here is [a pastebin](https://pastebin.com/0qkhyw1V).

zkochan:
are symlinks preserved during this caching? I think there are some symlinks in `.pnpm-store` and archives sometimes break symlinks

qm3ster:
Symlinks from inside `.pnpm-store` to inside `.pnpm-store`?
That does sound like a plausible cause.
@markoa do you know if they are preserved?

markoa:
I had to ask the team, the answer is that Semaphore cache currently doesn't
preserve symlinks.

PR #38, opened 2019-04-29
docs(hooks): Add link to `@pnpm/hook‑utils`

ExE-Boss:
This adds a link to the newly created [`@pnpm/hook-utils`](https://www.npmjs.com/package/@pnpm/hook-utils) package.

review?(@zkochan)

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -13,6 +13,9 @@ Hooks can be declared in a file called `pnpmfile.js`. `pnpmfile.js` should live 
 |`hooks.readPackage(pkg, context): pkg` | Allows to mutate every dependency's `package.json` |
 |`hooks.afterAllResolved(shrinkwrap, context): shrinkwrap` | Is called after resolution stage. Allows to mutate the shrinkwrap object. |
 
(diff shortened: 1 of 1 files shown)


docs/hooks.md
```diff
@@ -13,6 +13,9 @@ Hooks can be declared in a file called `pnpmfile.js`. `pnpmfile.js` should live
 |`hooks.readPackage(pkg, context): pkg` | Allows to mutate every dependency's `package.json` |
 |`hooks.afterAllResolved(shrinkwrap, context): shrinkwrap` | Is called after resolution stage. Allows to mutate the shrinkwrap object. |
 
+It's recommended to use the [`@pnpm/hook-utils`](https://www.npmjs.com/package/@pnpm/hook-utils)
```
zkochan: I don't think I would recommend it.

* it is good only for better logging.
* it has no documentation currently
* I don't know how any dependency in pnpmfile may be used. `@pnpm/hook-utils` should be installed globally? But in that case, the package may be broken.

TBH, I don't want to promote hooks too much. We should come up with declarative pnpmfile  anyway https://github.com/pnpm/pnpm/issues/1082

ExE-Boss: `@pnpm/hook-utils` is installed in whichever package has a `pnpmfile.js`.

PR #39, opened 2019-05-20
Small enhancements

lenovouser:


zkochan:
Are the demo animations still working after the optimization? I turned off the docusaurus optimizations because they were breaking the demo animations

lenovouser:
@zkochan Ugh, I thought about that when running svgo but forgot to check. You're right, they break when optimizing. Removed them.

PR #46, opened 2019-08-02
docs: enable search

juanpicado:
Things we need:

* API KEY
* About the **indexName** not pretty sure

* https://docusaurus.io/docs/en/search#docsNav

API_Keys___Algolia


zkochan:
I see, now the process is simpler. I submitter pnpm. Waiting for the key

PR #48, opened 2019-08-27
fix: correct linking for 'getting started' page

### LINKED ISSUE #47: "Getting Started" routes to incorrect url (homepage & footer)
From the homepage, clicking on the "getting started" button routes to [https://pnpm.js.org/docs/en/installation.html](https://pnpm.js.org/docs/en/installation.html) instead of [https://pnpm.js.org/en/installation](https://pnpm.js.org/en/installation) because it's using `docUrl(...)` to build the href path as seen below (same behavior with link in footer).

https://github.com/pnpm/pnpm.github.io/blob/c04475497d68de4653895aba7b09d0c6c9515f12/website/pages/en/index.js#L78

Pretty sure these should be `pageUrl(...)` instead.

Connorh999:
Fixes #47 

Ran docusaurus locally and everything checks out.

### DIFF
--- a/website/core/Footer.js
+++ b/website/core/Footer.js
@@ -8,18 +8,12 @@
 const React = require('react');
 
 class Footer extends React.Component {
-  docUrl(doc, language) {
-    const baseUrl = this.props.config.baseUrl;
-    return baseUrl + 'docs/' + (language ? language + '/' : '') + doc;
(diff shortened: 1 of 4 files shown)


zkochan:
Thanks
I'll publish it this evening

PR #49, opened 2019-08-29
Fix links in documentation

clottman:
The links don't work when they have the .md ending - removing the .md fixes the links.


zkochan:
Thanks! I'll publish this tonight

PR #58, opened 2020-01-08
Fix only-allow-pnpm example

Jolg42:
It looks like the "`" back tick makes yarn and pnpm hang.

Removing them make yarn console the message and pnpm works like expected!

### DIFF
--- a/docs/only-allow-pnpm.md
+++ b/docs/only-allow-pnpm.md
@@ -9,7 +9,7 @@ To prevent devs from using other package managers, add the following `preinstall
 ```json
 {
   "scripts": {
-    "preinstall": "node -e \"!process.env.npm_config_user_agent.startsWith('pnpm/') && !console.log('Use \\`npx pnpm install\\` to install dependencies in this repository\\n')&&process.exit(1)\""
+    "preinstall": "node -e \"!process.env.npm_config_user_agent.startsWith('pnpm/') && !console.log('Use \\`npx pnpm install\\` to install dependencies in this repository\\n') && process.exit(1)\""
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/only-allow-pnpm.md
```diff
@@ -9,7 +9,7 @@ To prevent devs from using other package managers, add the following `preinstall
 ```json
 {
   "scripts": {
-    "preinstall": "node -e \"!process.env.npm_config_user_agent.startsWith('pnpm/')&&!console.log('Use `npx pnpm install` to install dependencies in this repository\\n')&&process.exit(1)\""
```
zkochan: I think it would be better to still quote it. Maybe through `\\'`

Jolg42: I agree, doing it now :)

docs/only-allow-pnpm.md
```diff
@@ -9,7 +9,7 @@ To prevent devs from using other package managers, add the following `preinstall
 ```json
 {
   "scripts": {
-    "preinstall": "node -e \"!process.env.npm_config_user_agent.startsWith('pnpm/')&&!console.log('Use npx pnpm install to install dependencies in this repository\\n')&&process.exit(1)\""
+    "preinstall": "node -e \"!process.env.npm_config_user_agent.startsWith('pnpm/') && !console.log('Use \\`npx pnpm install\\` to install dependencies in this repository\\n')&&process.exit(1)\""
```
zkochan: I thought the ` had issues so you'll use the single quote

Jolg42: So yes indeed but it looks like the double escape works with the ` after testing it :)

Jolg42:
I just added 2 spaces to make it more readable but this version works for me perfectly! Thank you for the example by the way, because I messed up the dependencies with yarn once and didn't want that to happen again 😅

PR #60, opened 2020-03-10
docs: fix typo

nikoladev:
As reported by @paxperscientiam in https://github.com/pnpm/pnpm/issues/2409

### DIFF
--- a/docs/aliases.md
+++ b/docs/aliases.md
@@ -14,7 +14,7 @@ With aliases, you have a third option.
 Publish a new package called `awesome-lodash` and install it using `lodash` as its alias:
 
 ```
-pnpm add lodash@npm:awesome-lodash
+pnpm add awesome-lodash@npm:lodash
 ```
 
 No changes in code are needed. All the requires of `lodash` will import `awesome-lodash`.


docs/aliases.md
```diff
@@ -14,7 +14,7 @@ With aliases, you have a third option.
 Publish a new package called `awesome-lodash` and install it using `lodash` as its alias:
 
 ```
-pnpm add lodash@npm:awesome-lodash
+pnpm add awesome-lodash@npm:lodash
```
zkochan: In this case, it is not a typo because if you read the message above, it says:

```
Publish a new package called `awesome-lodash` and install it using `lodash` as its alias:
```

so awesome-lodash is the package in the registry.

But we can create a different example that will work in the console

nikoladev: Ah, I see! If I'm reading the preceding paragraph correctly, for the different example we would need 2 packages that (a) are published and (b) have the same API for this to not only work in the console but also make sense. Do you know any packages like this?

If not, I think we can just close this PR as it seems that this is a very rare problem and the original text does a good enough job of explaining how aliases work

PR #61, opened 2020-03-14
Add SwissDevJobs.ch on the users page

Vrq:


Vrq:
@zkochan @markoa could you have a look?

website/siteConfig.js
```diff
@@ -85,6 +85,13 @@ const users = [
     pinned: true,
     width: 250,
   },
+  {
+    caption: 'SwissDev Jobs',
+    image: '/img/users/swissdev-javascript-jobs.svg',
+    infoLink: 'https://swissdevjobs.ch/jobs/JavaScript/All/',
```
zkochan: I guess, add a link to the homepage because this way it looks like an ad

Vrq: We thought that this will be more relevant because our tech stack is fully JavaScript (node, react, pnpm) but if you like I can also change to the main page.

zkochan:
I'll update the website tonight

PR #62, opened 2020-03-15
added caching config of pnpm-store for travisci

rstiller:


docs/continuous-integration.md
```diff
@@ -12,8 +12,11 @@ On [Travis CI](https://travis-ci.org/), you can use pnpm for installing your dep
 ```yaml
 cache:
   npm: false
+  directories:
+    - "~/.pnpm-store"
 before_install:
   - curl -L https://unpkg.com/@pnpm/self-installer | node
+  - npm config set store-dir ~/.pnpm-store
```
zkochan: but that is the default location of the store. Why explicitly setting it?

rstiller: Just to remind people, who specify a custom location, to properly set it as well.

zkochan: ok, use `pnpm config` instead of `npm config`

rstiller: Done

PR #69, opened 2020-05-15
Add TakeShape to user list

mcat:
We just moved away from yarn and couldn't be more excited! We're publishing in detailed blog post about our migration next week.


zkochan:
Great news! Have you tried the alpha version of pnpm v5? It is going to be awesome!

mcat:
@zkochan we just posted the article on our switch from Yarn to pnpm - https://www.takeshape.io/articles/why-we-switched-from-yarn-to-pnpm/

zkochan:
So you don't use a shared lockfile? I'd recommend to use a shared lockfile in the monorepo. pnpm works a lot faster with a single lockfile.

I understand that `yarn <cmd>` is pretty convenient. Maybe we'll support that as well. However, to me it looks pretty weird as the main feature of a package manager is installing dependencies. So I'd expect `yarn something` to install `something`.

mcat:
We do use a single lockfile. We'll add a few sentences to clarify that in the article.

zkochan:
I see. That's great.

PR #70, opened 2020-05-18
Package store documentation out of sync

sean9keenan:
I believe after inspecting both the file generated, and the source here: https://github.com/pnpm/pnpm/blob/v4.14.3/packages/types/src/store.ts

It seems the store json is actually the inverse of what's documented


zkochan:
that file will be removed in pnpm v5

PR #76, opened 2020-08-21
Put actually working docker image in docs

karelbilek:
Note - it is 6 month old version of pnpm, and seems entirely unofficial and with just 1 build in february.

Probably shouldn't be even mentioned in the docs?


karelbilek:
https://github.com/znck/docker-pnpm/issues/1

zkochan:
I guess we should have an official image and publish it after publishing  the package

PR #77, opened 2020-09-08
Add tinyhttp to the list of projects using pnpm

talentlessguy:
tinyhttp uses pnpm for monorepo management, to be more specific:

- [worskpaces](https://github.com/talentlessguy/tinyhttp/blob/master/pnpm-workspace.yaml)
- [dependency management](https://github.com/talentlessguy/tinyhttp/blob/master/package.json)

tinyhttp has 250+ stars so I think it's ok to add it :D


talentlessguy:
yaaay

PR #84, opened 2020-12-22
Docs update: Add prezi to users

delanni:
It's a long overdue task I should have done :) 

Prezi is a proud and happy user of PNPM, so we're happy to show our faces under the users list.

- also fixing artsolution.de's url, it was missing a `/` in the url


zkochan:
Thanks, we're glad to see Prezi in our users list!

PR #88, opened 2021-02-14
docs(*): cleanup and rewrite for v5

Nytelife26:


Nytelife26:
Note that I am still working on this, it just is not my highest priority at the moment.

zkochan:
It would be better to make many small pull requests. It is hard to review so many changes.

also, this should be finished soon because I want to try to migrate to docusaurus 2

Nytelife26:
@zkochan It wouldn't, the pull request handles one scoped change, and overall the workload is the same whether it is in one or many. And, yes, it will be finished today - although, it won't prevent you from doing so whether it is or isn't.

docs/about-the-package-store.md
```diff
@@ -0,0 +1,55 @@
+---
```
zkochan, line 1: this page can be removed because the global store is not as described here at all. The global store is a content-addressable store.

Nytelife26: Resolved in latest commit.

Nytelife26:
@zkochan This is now ready to my knowledge. Let me know if you want me to squash the branch commits, or submit a change request if anything needs to be resolved.

docs/cli/run.md
```diff
@@ -5,35 +5,41 @@ title: pnpm run
...
 
-Runs a defined package script.
-
-## Synopsis
-
-```text
-pnpm run [[-r] [--no-bail] [--no-sort] [--workspace-concurrency=&lt;number>]]
-     &lt;command> [-- &lt;args>...]
-```
+Runs a script defined in the package's metadata.
```
zkochan, line 8: Metadata might be confusing because it is how the bundle of all the package manifests is called as well. Probably better to say "package's manifest"
```suggestion
Runs a script defined in the package's manifest (`package.json`).
```

docs/cli/run.md
```diff
@@ -92,19 +102,16 @@ Added in: v5.1.0
 
 Completely disregard concurrency and topological sorting, running a given script
 immediately in all matching packages with prefixed streaming output. This is the
-preferred flag for long-running processes such as watch run over many packages.
-
-```text
-pnpm run --parallel watch
-```
+preferred flag for long-running processes over many packages, for instance, a
+lengthy build process.
```
zkochan, line 106: no, I don't think that a build process is a good example. This feature was added for watch tasks because they never stop. It is better to keep some concurrency for build tasks, otherwise, the system can potentially crash.

docs/cli/server.md
```diff
@@ -3,76 +3,88 @@ id: server
...
 
 * Default: **false**
 * Type: **Boolean**
 
-Disallows stopping the server using `pnpm server stop`.
+Set whether to make the package store immutable to external processes while
+the server is running or not.
 
-### --ignore-upload-requests
+### --store
```
zkochan, line 72: I believe this was removed to `store-dir`, which is already documented above.

docs/cli/stop.md
```diff
@@ -5,10 +5,8 @@ title: pnpm stop
```
zkochan, line 5: since the `run` part is optional with any script, this page probably doesn't make sense anymore

docs/cli/store.md
```diff
@@ -7,43 +7,33 @@ Managing the [package store](../about-package-store).
...
-```
-
-Adds new packages to the pnpm store directly. 
-Does not modify any projects or files outside the store.
-
-Usage examples:
-
-```sh
-pnpm store add express@4 typescript@2
-```
```
zkochan, line 35: why was the example removed?

docs/cli/test.md
```diff
@@ -5,23 +5,8 @@ title: pnpm test
...
-```
-
-## Options
-
-### --recursive, -r
-
-Run the tests in every package found in subdirectories
-or every workspace package, when executed inside a workspace.
-
-### --filter &lt;package_selector>
```
zkochan, line 25: I think it is useful to have these options duplicated here as well. Or at least a link to the run command that contains them, which you also removed.

docs/completion.md
```diff
@@ -5,18 +5,23 @@ title: Command line tab-completion
 
 Added in: v4.9.0
 
-pnpm supports command line tab-completion for Bash, Zsh, and Fish shells.
+Unlike other popular package managers, which usually require plugins, pnpm
+supports command line tab-completion for Bash, Zsh, Fish, and similar shells.
```
zkochan, line 9: currently, we only support those three, so I guess "similar shells" is redundant. PowerShell is not supported at the momen.t

docs/continuous-integration.md
```diff
@@ -6,16 +6,19 @@ title: Continuous Integration
...
 ## Tips
-### Speeding-up
-You can speed things up by:
+
+### Improving performance
+
 ##### - Disable store signature verification
+
 ```shell
 pnpm set verify-store-integrity false
```
zkochan, line 15: this option doesn't make a difference since we use a content-addressable store, so this section may be removed.

docs/faq.md
```diff
@@ -5,107 +5,140 @@ title: Frequently Asked Questions
...
-Using symlinks on Windows is problematic indeed. That is why pnpm uses junctions instead of symlinks on Windows OS.
+[junctions]: https://docs.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions
 
-## Does it work on Windows? Nested `node_modules` approach is basically incompatible with Windows
+## But the nested `node_modules` approach is incompatible with Windows?
 
-Early versions of npm had issues because of nesting all `node_modules` (see [Node's nested node_modules approach is basically incompatible with Windows](https://github.com/nodejs/node-v0.x-archive/issues/6960)). However, pnpm does not create deep folders, it stores all packages flatly and uses symlinks to create the dependency tree structure.
+Early versions of npm had issues because of nesting all `node_modules` (see
+[this issue]. However, pnpm does not create deep folders, it stores all packages
+flatly and uses hard links to create the dependency tree structure.
```
zkochan, line 36: no, it uses symlinks to create the dependency tree structure.

With hard links, pnpm adds the files to the `node_modules/.pnpm` directory but symlinks are used to interconnect the dependencies.

```suggestion
flatly and uses symlinks to create the dependency tree structure.
```

docs/faq.md
```diff
@@ -5,107 +5,140 @@ title: Frequently Asked Questions
...
 One package can have different sets of dependencies on one machine.
 
-In project **A** `foo@1.0.0` can have dependency resolved to `bar@1.0.0` but in project **B** the same dependency of `foo` might resolve to `bar@1.1.0`. So pnpm hard links `foo@1.0.0` to every project where it is used, in order to create different sets of dependencies for it.
+In project **A** `foo@1.0.0` can have a dependency resolved to `bar@1.0.0`, but
+in project **B** the same dependency of `foo` might resolve to `bar@1.1.0`; so,
+pnpm hard links `foo@1.0.0` to every project where it is used, in order to
+create different sets of dependencies for it.
+
+Direct symlinking to the global store would work with Node's
+`--preserve-symlinks` flag, however, that approach comes with a plethora of its
```
zkochan, line 58: "plethora" I learned a new word today😀

docs/faq.md
```diff
@@ -5,107 +5,140 @@ title: Frequently Asked Questions
...
 
-If the store path is not set, then multiple stores are created (one per each drive or filesystem).
+If the store path is not set, then multiple stores are created (one per drive or
+filesystem).
 
-If installation is run on disk `D:`, the store will be created in `D:\.pnpm-store`. 
-If later the installation is run on disk `C:`, an independent store will be created in `C:\.pnpm-store`.
-The projects would still maintain the benefits of pnpm, but each drive may have redundant packages.
+If installation is run on disk `A`, the store will be created on `A` at
+`.pnpm-store`.  If later the installation is run on disk `B`, an independent
```
zkochan, line 91: I would still note somehow that the store will be at the root of the disk
```suggestion
`<disk A root>/.pnpm-store`.  If later the installation is run on disk `B`, an independent
```

docs/faq.md
```diff
@@ -5,107 +5,140 @@ title: Frequently Asked Questions
...
+its own list of deps.
 
-You can do so, by installing it via:
+The easiest solution to resolve missing dependencies of the buggy packages is to
+**add `iterall` as a dependency to our project's `package.json`**.
 
-`pnpm i iterall`
-
-It will be automatically added to your project's `package.json`.
+You can do so, by installing it via `pnpm i iterall`, and will be automatically
```
zkochan, line 140:
```suggestion
You can do so, by installing it via `pnpm add iterall`, and will be automatically
```

docs/faq.md
```diff
@@ -117,40 +150,50 @@ It will be automatically added to your project's `package.json`.
...
-An example was [Webpack Dashboard](https://github.com/pnpm/pnpm/issues/1043) which wasn't working with `pnpm`. It has since been resolved such that it works with `pnpm` now.
+An example was [Webpack Dashboard] which wasn't working with `pnpm`. It has
+since been resolved such that it works with `pnpm` now.
 
 It used to throw an error:
 
 ```console
 Error: Cannot find module 'babel-traverse'
-  at /node_modules/.registry.npmjs.org/inspectpack/2.2.3/node_modules/inspectpack/lib/actions/parse
+  at /node_modules/inspectpack/2.2.3/node_modules/inspectpack/lib/actions/parse
```
zkochan, line 163:
```suggestion
  at /node_modules/inspectpack@2.2.3/node_modules/inspectpack/lib/actions/parse
```

docs/faq.md
```diff
@@ -117,40 +150,50 @@ It will be automatically added to your project's `package.json`.
...
+    readPackage: (pkg) => {
+      if (pkg.name === "inspectpack") {
+        pkg.dependencies['babel-traverse'] = '^6.26.0';
       }
-      return pkg
+      return pkg;
     }
   }
-}
+};
```
zkochan, line 183: we use the Standard style in the pnpm codebase. Though I don't care that much about this example.

docs/faq.md
```diff
@@ -117,40 +150,50 @@ It will be automatically added to your project's `package.json`.
...
+
+[Webpack Dashboard]: https://github.com/pnpm/pnpm/issues/1043
 
 ### Solution 3
 
-In case there are too many issues, you can use the `shamefully-hoist` config. This creates a flat `node_modules` structure similar to the one created by `npm` or `yarn`.
+In case there are too many issues, you can use the `shamefully-hoist` option.
+This creates a flat `node_modules` structure similar to the one created by `npm`
+or `yarn`, which is not recommended as avoiding this structure is the primary
+purpose of pnpm.
```
zkochan, line 197: I'd be more specific because pnpm has many purposes. I am not sure which one is the primary.

```suggestion
purpose of pnpm's unique node_modules structure.
```

docs/filtering.md
```diff
@@ -5,126 +5,131 @@ title: Filtering
...
 
-## --filter &lt;package_name>
+## --filter \<package_name>
 
 Added in: v2.13.0
 
-To select an exact package, just specify its name (`@babel/core`) or use a pattern
-to select a set of packages (`@babel/*`).
+To select an exact package, just specify its name (`@scope/pkg`) or use a
+pattern to select a set of packages (`@scope/*`).
```
zkochan, line 26: it is recommended to quote args with *

```suggestion
pattern to select a set of packages (`"@scope/*"`).
```

docs/limitations.md
```diff
@@ -3,17 +3,24 @@ id: limitations
...
-npm's lockfile format. However, see [pnpm import](cli/import).
-2. You can't publish npm modules with `bundleDependencies` managed by pnpm.
-3. Binstubs (files in `node_modules/.bin`) are always shell files not
+1. `npm-shrinkwrap.json` and `package-lock.json` are ignored. Unlike pnpm, npm
+can install the same `name@version` multiple times and with different sets of
+dependencies. npm's lockfile is designed to reflect the flat `node_modules`
+layout, however, as pnpm cannot create a similar layout, it cannot respect
+npm's lockfile format. See [pnpm import] if you wish to convert a lockfile to
+pnpm's format, though.
+2. pnpm can't publish npm packages with `bundledDependencies` yet.
```
zkochan, line 12: I don't think we will ever support bundledDependencies. They don't work very consistently in npm as well. And you can always bundle your dependencies with a bundler instead of using bundledDependencies.
```suggestion
2. pnpm can't publish npm packages with `bundledDependencies`.
```

zkochan (changes requested):

zkochan:
so many changes, still not finished reviewing all.

zkochan:
If you have a PayPal, I can transfer you some appreciation from our OpenCollective funds.

docs/symlinked-node-modules-structure.md
```diff
@@ -3,84 +3,114 @@ id: symlinked-node-modules-structure
...
 
-These are the only "real files" in `node_modules`. Once all the packages are hard linked to `node_modules`, symlinks are
-created to build the nested dependency graph structure.
+These are the only "real" files in `node_modules`. Once all the packages are
+hard linked to `node_modules`, symbolic links are created to build the nested
+dependency graph structure.
 
-As you might have noticed, both packages are hard linked into a subfolder inside a `node_modules` folder (`foo@1.0.0/node_modules/foo`).
-This is needed to:
+As you might have noticed, both packages are hard linked into a subfolder inside
```
zkochan, line 36: this might be better
```suggestion
As you might have noticed, files of both packages are hard linked into a subfolder inside
```

docs/symlinked-node-modules-structure.md
```diff
@@ -3,84 +3,114 @@ id: symlinked-node-modules-structure
...
-      └─ bar -> ../../../bar@1.0.0/node_modules/bar
+|-- foo -> ./.pnpm/foo@1.0.0/node_modules/foo
+`-- .pnpm
+    |-- bar@1.0.0
+    |   `-- node_modules
+    |       `-- bar -> <store>/bar
+    `-- foo@1.0.0
+        `-- node_modules
+            |-- foo -> <store>/foo
+            `-- bar -> ../../../bar@1.0.0/node_modules/bar
```
zkochan, line 74:
```suggestion
            `-- bar -> ../../bar@1.0.0/node_modules/bar
```

docs/symlinked-node-modules-structure.md
```diff
@@ -3,84 +3,114 @@ id: symlinked-node-modules-structure
...
-      ├─ foo
-      └─ bar -> ../../../bar@1.0.0/node_modules/bar
+`-- .pnpm
+    |-- bar@1.0.0
+    |   `-- node_modules
+    |       `-- bar -> <store>/bar
+    `-- foo@1.0.0
+        `-- node_modules
+            |-- foo -> <store>/foo
+            `-- bar -> ../../../bar@1.0.0/node_modules/bar
```
zkochan, line 58:
```suggestion
            `-- bar -> ../../bar@1.0.0/node_modules/bar
```

docs/workspaces.md
```diff
@@ -3,93 +3,119 @@ id: workspaces
...
-In that case, pnpm will only link packages from the workspace if the `workspace:` protocol is used.
+[link-workspace-packages]: #link-workspace-packages
 
 ### Referencing workspace packages through aliases
 
 Added in 5.12.0
 
-For instance, the package in the workspace may be named `foo`. Usually, you would reference it as `{ "foo": "workspace:*" }`.
+Let's say you have a package in the workspace named `foo`. Usually, you would
+reference it as `"foo": "workspace:foo"`.
```
zkochan, line 43:
```suggestion
reference it as `"foo": "workspace:*"`.
```

docs/workspaces.md
```diff
@@ -3,93 +3,119 @@ id: workspaces
...
 
-If you want to use a different alias, the next syntax will work: `{ "bar": "workspace:foo@*" }`.
+If you want to use a different alias, the following syntax will work too:
+`"bar": "workspace:foo@*"`.
 
-Before publish, aliases are converted to regular aliased dependencies. The above example will become: `{ "bar": "npm:foo@1.0.0" }`.
+Before publish, aliases are converted to regular aliased dependencies. The above
+example will become: `"bar": "npm:foo@1.0.0"`.
```
zkochan, line 49: I am not sure looks better without the curly brackets

docs/workspaces.md
```diff
@@ -129,34 +141,43 @@ Added in: v5.13.0
...
-A shared lockfile also means that all dependencies of all workspace packages will be in a single `node_modules`.
+If this is enabled, pnpm creates a single `pnpm-lock.yaml` file in the root of
+the workspace. This also means that all dependencies of workspace packages will
+be in a single `node_modules` (and get symlinked to their package `node_modules`
+folder for Node's module resolution).
 
 Advantages of this option:
-
-* every dependency is a singleton
+* every dependency is a standalone
```
zkochan, line 164: I am not sure standalone and singleton are the same. It is important to note that it is a singleton because it is important that peer dependencies are singletons in a monorepo

zkochan (changes requested):

docs/cli/outdated.md
```diff
@@ -3,34 +3,23 @@ id: outdated
...
 
-## Synopsis
-
-```text
-pnpm outdated [-r] [--filter &lt;package selector>]
-              [&lt;package pattern> ...]
-
-pnpm recursive outdated [--filter &lt;package selector>]
-                        [&lt;package pattern> ...]
-```
```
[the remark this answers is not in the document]
Nytelife26: It adds unnecessary bulk to the file, would be a pain to clean up because of all the unnecessary escaped HTML that shouldn't be in a code block, and is less helpful than running `pnpm <cmd> --help` in the terminal as well as providing no additional value over the documentation itself as a whole and the example tables found in TL;DR sections.

docs/cli/run.md
Nytelife26: Build processes are almost always run in parallel. Albeit usually with some degree of constraint, they are. Monorepositories for instance. Sapphire has over 35 packages and builds them all in parallel because, well, could you imagine if they didn't? 

The example isn't the important part. What it demonstrates is.

docs/cli/recursive.md
```diff
@@ -3,30 +3,32 @@ id: recursive
...
-|--|--|
-|`pnpm install -r` | runs installation for every package in every subfolder |
-|`pnpm run build --filter foo-*` |builds all packages with names that start with `foo-` |
-|`pnpm update -- login-page...` |updates dependencies in `login-page` and any dependencies of `login-page` that are also in the repository |
+Aliases: `m`, `multi`, `<command> -r`
 
 ## Options
 
-### --filter &lt;package_selector>
+### --link-workspace-packages
```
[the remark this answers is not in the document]
Nytelife26: It is, however, I thought it was still worth noting. Especially for cases where it may be disabled in npmrc and someone wishes to override that behaviour without having to change the npmrc.

docs/cli/server.md
Nytelife26: I have absolutely no idea. Both options are still available from CLI, so I'm uncertain as to whether either is deprecated. pnpm's development isn't very well documented. However, since they both exist in CLI, surely both should have a section?

docs/cli/stop.md
Nytelife26: It makes roughly as much sense as documenting `pnpm start` and the like. They're technically lifecycle scripts. Ultimately they don't need to be documented, and honestly, if I were any of the major package managers I wouldn't, but people do.

docs/cli/store.md
Nytelife26: It's a self-explanatory command. It doesn't require any examples because it does exactly as it says. I will however add a note to the documentation regarding its similarity to standard add, though.

docs/cli/test.md
Nytelife26: It does say it is an alias of `run test`, so logically speaking it inherits all options of `run`

docs/completion.md
Nytelife26: I wasn't referencing PowerShell. Rather, the plethora of other *NIX shells that are essentially functionally equivalent to those main three. Exempli gratia, csh, ash, mksh, pdksh, dash, et cetera.

docs/faq.md
Nytelife26: It will be at the root of the filesystem or partition, not at the root of the disk, however I will add that note.

docs/faq.md
Nytelife26: Glad I could be of assistance :)

docs/faq.md
Nytelife26: I'm glad you catch these things. Normally I have a keen eye for proofreading, but it's difficult to maintain for very large rewrites. Bravo :)

docs/faq.md
Nytelife26: That is standard style. It is universally agreed upon that semicolons should terminate statements in most programming languages since ASI can be very problematic.

docs/filtering.md
Nytelife26: That is demonstrated in the examples below. When writing generics we leave it down to the user to know what escaping their shell requires.

docs/npmrc.md
```diff
@@ -49,53 +56,63 @@ hoist-pattern[]=*babel*
...
+Added in: v1.34.0 as `shamefully-flatten`, renamed in v4.0.0
 
 * Default: **false**
 * Type: **Boolean**
 
-By default, pnpm creates a semistrict `node_modules`. It means that your dependencies have access to undeclared dependencies
-but your code does not. With this layout, most of the packages in the ecosystem work with no issues.
-However, if some tooling only works when the hoisted dependencies are in the root of `node_modules`, you can set this config to `true`. 
+By default, pnpm creates a semistrict `node_modules`, meaning dependencies have
+access to undeclared dependencies but main libraries themselves do not. With
```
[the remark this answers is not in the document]
Nytelife26: Isn't this false anyway? I'm fairly sure pnpm creates a fully strict layout, not semistrict, so even dependencies don't have access to them.

docs/package_json.md
```diff
@@ -21,41 +23,31 @@ You can specify the version of Node and pnpm that your stuff works on:
...
-            "optional": true
-        },
-        // express is not specified in the peerDependencies field,
-        // so pnpm will assume that any version of express is fine
-        "express": {
-            "optional": true
-        }
-    }
-}
-```
```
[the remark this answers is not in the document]
Nytelife26: Really? It's pretty self-explanatory. If optional is true, the peer dependency gets labelled as optional. Do people genuinely need examples for that? I'm not being condescending, just wondering.

docs/workspaces.md
Nytelife26: There is no need for them. It is a key of a JSON object, not a JSON object itself. Putting them in takes up space and looks clunky.

Nytelife26:
@zkochan I have left all changes with important discussion points unresolved, so feel free to gloss over those, and then I'll implement whatever verdict we reach. As for PayPal, I do indeed. It is not at all necessary so do not feel pushed to, but I would appreciate it :)

If you so wish, my personal payment page can be found at https://paypal.me/nytelife26

docs/cli/server.md
zkochan: `store` is kept for backward compatibility. `store-dir` is the new setting name. `store` should not be mentioned in the docs. The rename happened in pnpm v2 I believe

docs/cli/stop.md
zkochan: start makes sense to document because it has a default value `node server.js` (might be a good idea to change this in pnpm v6, to remove start as well)

docs/package_json.md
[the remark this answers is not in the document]
zkochan: I think they do, especially if their English is not very good. Also, the Yarn docs are more visual in this case https://yarnpkg.com/configuration/manifest

docs/releasing-apps.md
```diff
@@ -1,14 +0,0 @@
...
-
-There are two ways to release an app with pnpm. One way is to commit `pnpm-lock.yaml` into your repo.
-Which we recommend doing anyway. And then in prod you'll have just to run `pnpm install`.
-You'll be sure in that case that the same dependencies will be used, with which you tested your app in other environments.
-
-If you'd like to copy packages to prod, you'll have to commit `pnpm-lock.yaml` anyway. And you'll have to
-copy paste the global store to production. The global store location is configurable
-via the `store` config key.
-Then you can run `pnpm install --offline` in your app and pnpm will be using packages that are already in the
-global store without making any requests to the npm registry.
```
zkochan, line 14: why was this page removed?

docs/cli/stop.md
Nytelife26: Maybe it would be better overall to make a page regarding lifecycle events, similar to the one featured in npm's documentation? Either way, I'll remove stop for now.

docs/releasing-apps.md
Nytelife26: It wasn't, it got renamed to docs/production.md since not all projects are applications as such but all projects operate in a production environment, which is what the page is really about.

Nytelife26:
I have implemented all the requested changes. This should be ready for merge, however, I will first check that all hyperlinking is valid and report back.

Nytelife26:
@zkochan All requested changes have been added. This is now ready to be merged as far as I can tell. As an aside though, I will update the readme with some necessary build steps I encountered that are undocumented.

zkochan (approved):

zkochan:
@Nytelife26 you should get an email from OpenCollective.

Nytelife26:
@zkochan I have indeed. That's more than I was expecting, thank you very much! This is actually the first non-contract OSS project I've been paid for too, so much appreciated :) I just submitted the invoice as requested in the email.

Have a wonderful day.

PR #93, opened 2021-03-11
Update continuous-integration.md

dominikg:
be more specific in sudo note

### DIFF
--- a/versioned_docs/version-5.x/continuous-integration.md
+++ b/versioned_docs/version-5.x/continuous-integration.md
@@ -116,6 +116,6 @@ jobs:
 
 :::note
 
-Using `actions/setup-node@v2` you need to install with [root permissions](https://github.com/actions/setup-node/issues/177), eg:`sudo `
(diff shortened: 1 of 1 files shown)


versioned_docs/version-5.x/continuous-integration.md
```diff
@@ -116,6 +116,6 @@ jobs:
 
 :::note
 
-Using `actions/setup-node@v2` you need to install with [root permissions](https://github.com/actions/setup-node/issues/177), eg:`sudo `
+Using `actions/setup-node@v2` you need to install pnpm with [root permissions](https://github.com/actions/setup-node/issues/177), eg:`sudo ` if you do not specify a node version to use
```
zkochan, line 119: I would write it in a separate sentence

```suggestion
Using `actions/setup-node@v2` you need to install pnpm with [root permissions](https://github.com/actions/setup-node/issues/177), eg:`sudo npm install -g pnpm`. Alternatively, if you specify the Node.js version to use, pnpm may be installed with no priviledged user.
```

PR #95, opened 2021-03-27
migrate benchmark repo

juanpicado:
- Migrate https://github.com/pnpm/benchmarks-of-javascript-package-managers onto this repository
- Enable GitHub action to run Benchmarks
- Commit new benchmarks automatically (disabled for now)


package.json
```diff
@@ -6,6 +6,8 @@
     "publish-gh-pages": "docusaurus-publish --skip-image-compression",
     "write-translations": "rm -rf i18n/en && docusaurus write-translations",
     "version": "docusaurus docs:version",
+    "prebench:clean": "./node_modules/.bin/rimraf .tmp",
```
juanpicado, line 9: Not sure if there is a better way with pnpm.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,24 @@
+name: Benchmark runner
+
+on: push
```
juanpicado, line 3: currently running on every branch, I can limit this to master or schedule to run every day.

zkochan: I don't think it should run on push. It should run every day.

juanpicado: Set on 2:30am every day. Any preference?

zkochan (approved):

zkochan:
actually, I am not sure about this solution. I thought it would be a subproject with its own package.json. Now its dependencies are mixed with the website deps

juanpicado:
:) ok too late to change it, I can refactor it, maybe you can picture me the idea you have and follow up a PR

zkochan:
I might just move it around myself

zkochan:
moved

zkochan:
ok, I actually amended the change and force pushed now

juanpicado:
Nice touch. `pnpm --dir=benchmarks`, btw `--dir` is not at the docs? I cannot find it.

zkochan:
it is here
https://pnpm.js.org/pnpm-cli

PR #101, opened 2021-04-08
Set --no-audit & --legacy-peer-deps in benchmarks for npm

darcyclarke:
Setting these should more accurately compare performance as `yarn` & `pnpm` do not audit or install peer-deps by default.

### DIFF
--- a/benchmarks/commandsMap.js
+++ b/benchmarks/commandsMap.js
@@ -14,7 +14,6 @@ module.exports = {
       'cache',
       '--registry',
       'https://registry.npmjs.org/',
-      '--no-audit',
       '--legacy-peer-deps'
     ]
   },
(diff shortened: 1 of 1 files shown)


zkochan:
`--no-audit` is already there. And I am not sure adding `--legacy-peer-deps` is fair. We try to compare the package managers using their default settings. I don't think this option will make a difference anyway as I think we don't have missing peer dependencies in the fixtures

zkochan:
anyway, I will add it just because npm adds it in its benchmark. It will not make a big difference anyway https://github.com/npm/benchmarks/blob/b66f15c2378d966b7e718789e6c1339aeef1eb01/execute.js#L47

darcyclarke:
@zkochan totally gapped on `no-audit` being there already, had been looking at/reviewing the old benchmark repo so apologies there.

PR #103, opened 2021-04-18
Format commands as code

GreenRobot777:
For simplify translation process, for example [this](https://crowdin.com/translate/pnpm/2252/en-ru?filter=basic&value=0#q=Search)

### DIFF
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -22,7 +22,7 @@ pnpm <command> --filter <package_selector>
 
 :::
 
-## --filter &lt;package_name>
+## `--filter <package_name>`
 
 Added in: v2.13.0
 
@@ -37,9 +37,9 @@ pnpm test --filter "@babel/*"
 pnpm test --filter "*core"
 ```
 
-## --filter &lt;package_name>...
+## `--filter <package_name>...`
 
-Added in: v2.13.0
+Added in: `v2.13.0`
 
 To select a package and its dependencies (direct and non-direct), suffix the
 package name with an ellipsis: `<package_name>...`. For instance, the next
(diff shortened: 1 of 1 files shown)


zkochan:
I am not sure about this because it changes the way it is rendered.

zkochan:
maybe we could capitalize instead.

what do you think @pnpm/collaborators ?

```
# --filter=PACKAGE_NAME
```

shellscape:
capitalizing is good by me

Aankhen:
These aren’t code, though. They should be marked up with <kbd>&lt;kbd&gt;</kbd>.

zkochan:
Looks like this with \<kbd>:

image

I think it is fine just to remove the \< and >. Even capitalizing is not required.

Aankhen:
I can’t comment on what what be most useful for the translation, but to format this correctly you’d need to use <kbd>&lt;kbd&gt;</kbd> and <kbd>&lt;var&gt;</kbd> together:
```markdown
<kbd>--filter <var>package_name</var></kbd>
```
Which would give you:

<kbd>--filter <var>package_name</var></kbd>

I can understand if that’s a lot to repeat for every instance.

nikoladev:
@GreenRobot777 How do these changes simplify the translation process?

---

@Aankhen just now introduced me to `<kbd>` and `<var>` and it seems to be a very good way to use semantic HTML to give proper meaning. I think this would help people using assistive technology (like screenreaders).

As for how much work it would be to implement: I feel that the work can be reduced by having a template that the translators copy, paste, and edit. But I've never really done translating work so I can't say for sure.

---

> I don’t the version numbers should be marked up as either.

I agree with this.

PR #106, opened 2021-05-10
Add instruction for Bitbucket Pipelines

tomchentw:


docs/continuous-integration.md
```diff
@@ -142,3 +142,26 @@ build:
...
+    pnpm: $BITBUCKET_CLONE_DIR/.pnpm-store
+
+pipelines:
+  pull-requests:
+    "**":
+      - step:
+          name: Build and test
+          image: node:14.16.0
+          script:
+            - curl -f https://get.pnpm.io/v6.js | node - add --global pnpm
```
zkochan, line 162: It is better to install a given major version.

```suggestion
            - curl -f https://get.pnpm.io/v6.js | node - add --global pnpm@6
```

tomchentw: Done!

tomchentw:
I was wondering why this is not show up in the docs but ultimately realized it's the `next` version: https://pnpm.io/next/continuous-integration#bitbucket-pipelines

I will also add this to the v6.x.

PR #110, opened 2021-05-16
Feat/add yarn berry to benchmark

tomchentw:
## Description

This PR adds `yarn berry` into the benchmark. 

Previously, I mainly use `yarn berry` and I switched to `pnpm` recently which is really surprising. But I'm still curious about the performances differences though so here's the updated bench.

Feel free to modify anything!

## Prevent Global Cache

To get the correct bench result, one might want to remove global cache before running:

```
rm -rf ~/.pnpm-store
rm -rf ~/.yarn/berry
```

Note that after running the benchmark script, on my machine it will be:

```
> du -h -d=0 ~/.pnpm-store
34M	/Users/tomchentw/.pnpm-store
> du -h -d=0 ~/.yarn/berry
4.0K	/Users/tomchentw/.yarn/berry
```

Where the `34M` should be the dependencies for `npm`, `pnpm` and `yarn` during the initialization step.

## Screenshot

JPEG-1

### DIFF
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -1,5 +1,7 @@
 'use strict'
 import fs from 'fs'
+import { promisify } from 'util'
+import rimraf from 'rimraf'
 import mkdirp from 'mkdirp'
 import commonTags from 'common-tags'
 import prettyMs from 'pretty-ms'
@@ -11,6 +13,8 @@ import path from 'path'
 import { fileURLToPath } from 'url'
 
 const DIRNAME = path.dirname(fileURLToPath(import.meta.url))
+const TMP = path.join(DIRNAME, '.tmp')
+
 const { stripIndents } = commonTags
 const LIMIT_RUNS = 3
 
@@ -106,7 +110,10 @@ run()
 
 async function run () {
   const managersDir = path.join(DIRNAME, 'managers')
-  await fs.promises.mkdir(managersDir, { recursive: true })
+  await Promise.allSettled([
+    promisify(rimraf)(TMP),
+    fs.promises.mkdir(managersDir, { recursive: true }),
+  ])
   spawn.sync('pnpm', ['init', '--yes'], { cwd: managersDir })
   spawn.sync('pnpm', ['add', 'yarn@latest', 'npm@latest', 'pnpm@latest'], { cwd: managersDir, stdio: 'inherit' })
   const formattedNow = new Intl.DateTimeFormat('en-US', { dateStyle: 'medium', timeStyle: 'short' }).format(new Date())
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -2,7 +2,6 @@
 import fs from 'fs'
 import { promisify } from 'util'
 import rimraf from 'rimraf'
-import mkdirp from 'mkdirp'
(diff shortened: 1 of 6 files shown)


zkochan:
Why do we need to measure 3 versions of Yarn?

Let's remove Yarn v1 from the benchmarks. Let's test Yarn 2 with node-linker and Yarn 2 with PnP.
I am not even sure we need to measure Yarn 2 with PnP because it is like comparing apples to oranges.

tomchentw:
> Why do we need to measure 3 versions of Yarn?
> Let's remove Yarn v1 from the benchmarks

@zkochan Of course we could remove yarn v1 from the benchmarks but I would imagine there are plenty of yarn users still stay at v1 due to issues with v2. It's your call. No objections of removing it.

> Let's test Yarn 2 with node-linker and Yarn 2 with PnP.

From the docs, [yarn v2](https://yarnpkg.com/configuration/yarnrc#nodeLinker) uses the `pnp` mode by default so I would assume that's the majority option for yarn v2 users (and they promote this as a perf benefit!) I don't think we should test yarn v2 with the old node_modules strategy.

zkochan:
Well, if they want to see all the different Yarn versions, Yarn has its own benchmarks: https://yarnpkg.com/benchmarks
So I don't think we need to keep Yarn 1. But I would keep Yarn 2 with node-linked because it is similar to what pnpm does and a more fair comparison.

tomchentw:
That makes sense. PR is updated.

The results look really promising even compared with yarn berry PnP mode. Thank you for the great work!

## Screenshot

WithYarnBerry

benchmarks/index.js
```diff
@@ -100,15 +104,38 @@ const toArray = (pms, resultsObj) => {
...
+  const yarnPkgJson = JSON.parse(fs.readFileSync(yarnPkgJsonPath, 'utf-8'))
+  yarnPkgJson.version = yarnBerryVersion
+
+  // Replace yarn binary with the yarn berry script
+  await Promise.allSettled([
+    fs.promises.writeFile(yarnPkgJsonPath, JSON.stringify(yarnPkgJson)),
+    fs.promises.rename(path.join(DIRNAME, 'managers/.yarn/releases/yarn-berry.cjs'), path.join(DIRNAME, 'managers/node_modules/.bin/yarn')),
+    fs.promises.rm(path.join(DIRNAME, 'managers/.yarnrc.yml')),
+  ]);
+}
```
zkochan, line 122: it is insane how hard it is to configure Yarn Berry

zkochan (approved):

zkochan:
The version of Yarn is not correctly detected though

tomchentw:
What!? I thought I wrote the version into the package.json. No?

yeah it’s hard because they manage dependencies themselves 🤦

src/.tmp/npm/alotta-files/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "alotta-modules",
   "version": "0.0.1",
   "dependencies": {
-    "animate.less": "^2.2.0",
     "autoprefixer": "^6.0.3",
```
tomchentw, line 5: @zkochan ah I accidentally removed this. Sorry!

PR #119, opened 2021-06-22
fix typo inside `add.md`: `Exempli gratia` -> `For example`

TemaSM:
Related string/issue on Crowdin: https://crowdin.com/translate/pnpm/2976/en-ru#97282


zkochan (approved):
e.g. is fine as "for example" but ok

PR #129, opened 2021-07-28
Minor spelling fix

kitsune7:
Really not a huge deal, but since there's a convenient edit button on the site, I figured I could help out with a quick fix. In the context that it's being used, "affect" is the appropriate verb.

See [Affect vs. Effect](https://www.merriam-webster.com/words-at-play/affect-vs-effect-usage-difference) for more details.


zkochan:
thanks

PR #134, opened 2021-08-23
add Prisma to list of users

m-abdelwahab:
Tiny PR to include Prisma in the https://pnpm.io/users page :)

### DIFF
--- a/users.json
+++ b/users.json
@@ -7,6 +7,13 @@
     "width": 150
   },
   {
+    "caption": "Prisma",
+    "image": "/img/users/prisma.svg",
+    "infoLink": "https://prisma.io",
+    "pinned": true,
+    "width": 150
+  },
+  {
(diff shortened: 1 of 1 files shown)


zkochan:
but prisma is already in our sponsors section, which is above the users section.

zkochan:
image

I guess it is ok to put it to the separate page but then maybe set `pinned: false`, so that it is not duplicated on the homepage

users.json
```diff
@@ -6,6 +6,13 @@
     "pinned": true,
     "width": 150
   },
+  {
+    "caption": "Prisma",
+    "image": "/img/users/prisma.svg",
+    "infoLink": "https://prisma.io",
+    "pinned": true,
```
Jolg42, line 13:
```suggestion
    "pinned": false,
```

PR #144, opened 2021-09-26
Add telecraft as a pnpm workspace example

MKRhere:


zkochan:
How does it work though? I see that it uses `.pnpm-workspace.yaml` in the root of the repository but pnpm is only reading `pnpm-workspace.yaml`

MKRhere:
I'm not sure how this happened, must've been a mistake. pnpm probably defaulted to all packages of all subdirs. I've fixed it now.

PR #169, opened 2021-11-05
add podcast episode from syntax

mittalyashu:


juanpicado (approved):
LGTM

(@zkochan I'll check the PR failing later, secrets cannot be shared over forks)

PR #174, opened 2021-11-16
Rename users.json to user.json

Qdigital:


juanpicado:
Hi @Qdigital , what's the reason of this ?

PR #195, opened 2021-12-25
feat: Add yarn pnpm implementation to benchmarks

akronb:
In version 3.1 `yarn berry` added new install mode `pnpm` based on pnpm install strategy,

I added a new mode to benchmarks and made legend a bit wider since **Yarn pnpm** doesn't fit. 

I don't feel right about **Yarn pnpm** naming, if you have something better on your mind, tell me!

### DIFF
--- a/benchmarks/benchmarkFixture.js
+++ b/benchmarks/benchmarkFixture.js
@@ -85,6 +85,10 @@ export default async function benchmark (pm, fixture, opts) {
                 + 'nmMode: hardlinks-local\n'
                 + 'compressionLevel: 0\n'
         break
+      case 'yarn_pnpm':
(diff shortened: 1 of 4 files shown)


zkochan:
3 benchmarks for Yarn seems too much.

Could we just choose one or 2 modes of Yarn? We can link to Yarn's benchmarks that test all modes: https://yarnpkg.com/benchmarks

akronb:
Oh, I missed that benchmark. 

I agree about too many Yarn benchmarks. Looks like Yarn pnpm install mode doesn't do a big difference to node_modules mode, so I can close this pull request and create new with link to Yarn's benchmarks.

zkochan:
ok

PR #198, opened 2021-12-30
Fix smudged illustration

rob-lattice:
The first illustration on the motivation page looks smudged ... or something. This PR just replaces it with the original from Twitter.


zkochan:
The image is smudged again though didn't do anything.

rob-lattice:
Very strange. Is the Cloudfront distribution configured to compress or scale images ... somehow?

PR #208, opened 2022-01-12
fix: publish-branch needs to be prefixed

iChengbo:


zkochan:
but it may also be used as a setting in `.npmrc`:

```
publish-branch=somebranch
```

iChengbo:
> but it may also be used as a setting in `.npmrc`:
> 
> ```
> publish-branch=somebranch
> ```

@zkochan Yes, it is. But I think in the `docs/cli/publish.md` it refers to the publish command, not to configure it in `.npmrc`.

iChengbo:
Maybe this way should be added to https://github.com/pnpm/pnpm.github.io/blob/main/docs/npmrc.md.
```
publish-branch=somebranch
```

zkochan:
I don't think it is frequent to use it as an option. Maybe just move it to `.npmrc`

iChengbo:
Yes, it is infrequent, move it is also a good choice.

PR #216, opened 2022-02-03
docs(pnpm run): valid shell flags syntax

yurylavrukhin:
Is this the way `enable-pre-post-scripts` and `shell-emulator` flags/options should be written?
That's how the other flags on this page are written.

Or they are different entities and therefore written without `--` at the beginning?

Let me know if I am missing something 🙏🏻


zkochan:
They RC options that should be written to `.npmrc`.

PR #240, opened 2022-04-01
feat: add Agora Flat as user

BlackHole1:
This is open source project. 
See: https://github.com/netless-io/flat


users.json
```diff
@@ -281,5 +281,12 @@
     "infoLink": "https://element-plus.org/",
     "pinned": false,
     "width": 250
+  },
+  {
+    "caption": "Agora Flat",
+    "image": "/img/users/agora-flat.svg",
+    "infoLink": "https://flat.whiteboard.agora.io/en/",
+    "pinned": true,
```
juanpicado, line 289: Usually this is false, could you update?

BlackHole1: This is the expected modification 👀. From my personal point of view, I would expect the use of `pinned: true` here.
If you don't feel too good about it, please let me know.

juanpicado: I am fine with it, Zoltan has here the last word.

BlackHole1: @zkochan

zkochan: We usually pin only the most popular projects to the homepage.
Is Agora Flat well known?

BlackHole1: Okey 👌

PR #242, opened 2022-04-07
Update users.json

lassv:
Added treeprofile.com

### DIFF
--- a/users.json
+++ b/users.json
@@ -295,7 +295,7 @@
     "caption": "Treeprofile",
     "image": "https://www.treeprofile.com/assets/img/icon-green.png",
     "infoLink": "https://www.treeprofile.com",
-    "pinned": true,
+    "pinned": false,
     "width": 120
   }
 ]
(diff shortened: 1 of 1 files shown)


users.json
```diff
@@ -288,5 +289,13 @@
     "infoLink": "https://flat.whiteboard.agora.io/en/",
     "pinned": true,
     "width": 120
+  },
+  
+   {
+    "caption": "Treeprofile",
+    "image": "https://www.treeprofile.com/assets/img/icon-green.png",
+    "infoLink": "https://www.treeprofile.com",
```
juanpicado, line 297: What is this exactly?

lassv: @juanpicado My little cool website, where users can have their own profiles. And we are using **pnpm** for the project

users.json
```diff
@@ -288,5 +289,13 @@
...
+    "pinned": true,
```
juanpicado, line 298: This should be `false` by default, could you update it?

lassv: @juanpicado done

juanpicado (approved):
LGTM, thanks @lassv good luck with the initiative 💯

PR #250, opened 2022-04-20
mention that concurrent operations on the store are supported

tjenkinson:
refs https://github.com/pnpm/pnpm/issues/941#issuecomment-1103790251

### DIFF
--- a/docs/cli/store.md
+++ b/docs/cli/store.md
@@ -5,6 +5,8 @@ title: pnpm store
 
 Managing the package store.
 
+Concurrent operations on the store are supported, so it's safe to mount the store from a shared drive into multiple builds at the same time.
+
 ## Commands
 
 ### status


docs/cli/store.md
```diff
@@ -5,6 +5,8 @@ title: pnpm store
 
 Managing the package store.
 
+Concurrent operations on the store are supported, so it's safe to mount the store from a shared drive into multiple builds at the same time.
```
zkochan, line 8: But if you do this, pnpm will probably not hard link but copy files from the mounted drive. I don't think it is possible to create a hard link that points to a mounted drive. So installation will be slower and use more disk space.

tjenkinson: So the specific scenario we want to try is using CodeBuild with an efs mount on linux. I think even if pnpm copies from the mount to the actual package directory it still might be quicker than the alternative which is have pnpm start on an empty machine every time?

Shall I change this just to "Concurrent operations on the store are supported."?

PR #257, opened 2022-04-30
docs: Clarify how to check for latest version

thien-do:


thien-do:
Hm turn out this may not be expected as the "latest" currently is still version 6

juanpicado (approved):
LGMT I'm 99% here @zkochan

PR #273, opened 2022-06-08
docs: fix webpack run with option from script

ild0tt0re:


juanpicado:
Could you elaborate please? I don't manage to see why the change

zkochan (approved):

zkochan:
It was an error in the docs

PR #275, opened 2022-06-13
Avoid looking for packages at infinite depth.

nathanhammond:
As well-illustrated here, the documentation of `packages/**` will result in checking at arbitrary depths for `package.json` files in a misguided attempt to find workspaces. This isn't the way most humans would construct project layouts and is a performance footgun: https://github.com/vercel/turborepo/issues/667#issuecomment-1128057910


zkochan:
ok, but then I think it needs to be mentioned that there are two options `*` and `**`. There are projects that use multiple levels of directories.

nathanhammond:
Good to demonstrate all patterns that are available and this will hopefully encourage people to be more cognizant of their glob patterns. 👍

PR #276, opened 2022-06-14
docs(installation): adds Volta tool

ild0tt0re:
Adds `Volta` tool among the installation methods.

**Why?**
Probably the best **JavaScript Tool Manager**: https://docs.volta.sh/guide/

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -39,6 +39,14 @@ This will automatically install pnpm on your system. However, it probably won't 
 corepack prepare pnpm@<version> --activate
 ```
 
+## Using Volta
+
+If you have [Volta](https://volta.sh/) installed, you can install pnpm using the following command:
+
+```
+volta install pnpm
+```
+
 ## Using npm
 
 ```sh
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -29,13 +29,13 @@ You may use the [pnpm env] command then to install Node.js.
 
 Since v16.13, Node.js is shipping [Corepack](https://nodejs.org/api/corepack.html) for managing package managers. This is an experimental feature, so you need to enable it by running:
 
-```
+```sh
 corepack enable
 ```
 
 This will automatically install pnpm on your system. However, it probably won't be the latest version of pnpm. To upgrade it, check what is the [latest pnpm version](https://github.com/pnpm/pnpm/releases/latest) and run:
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -39,6 +39,14 @@ This will automatically install pnpm on your system. However, it probably won't
 corepack prepare pnpm@<version> --activate
 ```
 
+## Using Volta
+
+If you have [Volta](https://volta.sh/) installed, you can install pnpm using the following command:
+
+```
```
juanpicado, line 46: Could you add ```bash ? please

ild0tt0re: I added ```sh to be consistent to the other commands and also where missing

zkochan:
Since when does Volta support pnpm? I have not seen progress in their related issue.

ild0tt0re:
> Since when does Volta support pnpm? I have not seen progress in their related issue.

@zkochan you can use `Volta` to install every js bin tool you want.
The support that you ask for is relative to node.js and yarn where you can also pin a specific version to a repositories but this is an additional feature not needed to install binaries as you can see from my `volta ls` output:

```shell
⚡️ Currently active tools:

    Node: v16.13.1 (current @
    /Users/angeloca/dev/fe.monorepo/package.json)
    Yarn: v1.22.17 (current @
    /Users/angeloca/dev/fe.monorepo/package.json)
    Tool binaries available:
        yarn, yarnpkg (default)
        create-nx-workspace (default)
        dnotebook (default)
        jscodeshift (current @
        /Users/angeloca/dev/fe.monorepo/package.json)
        madge (default)
        nt, ntl, nodetask, nodetasklist, npm-tasklist
        (current @
        /Users/angeloca/dev/fe.monorepo/package.json)
        nx (current @
        /Users/angeloca/dev/fe.monorepo/package.json)
        pnpm, pnpx (default)
        yalc (default)
```

docs/installation.md
juanpicado: Thanks I'll clean up the others, on my list

zkochan:
Then I don't really see any reason listing Volta as an installation method. Volta doesn't really provide any features that pnpm doesn't provide. pnpm can also manage Node.js versions and when pnpm installs a package globally, it also pins it to the active node.js version.

PR #292, opened 2022-07-25
fix: deploy example dockerfile

zkochan:
ref pnpm/pnpm#5082

### DIFF
--- a/docs/cli/deploy.md
+++ b/docs/cli/deploy.md
@@ -32,9 +32,8 @@ WORKDIR /app
 
 ENV NODE_ENV=production
 
-COPY --from=pruned /app/pruned/dist dist
-COPY --from=pruned /app/pruned/node_modules node_modules
+COPY --from=pruned /app/pruned .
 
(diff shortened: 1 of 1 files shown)


juanpicado (approved):
This is how actually should be. 👍

shellscape:
There are a number of issues with the example, even as it is in this PR:

- The `Dockerfile` must be run from the repo root (that's not mentioned)
- `FROM workspace AS pruned` assumes that there's a previous image named `workspace` for that `FROM` to pull from, and it assumes that `pnpm` has been installed
- The `FROM node:18-alpine` image doesn't install `pnpm` and will fail on `RUN ...`
- The example is missing some critical steps, such as copying the monorepo into the image in order to install and fetch, etc. - If the Dockerfile is within the directory of a workspace package/service, and `docker build` is run within that directory, it won't be able to see the repo root file contents

So while `pnpm deploy` is a very nice command, the docs around it are not exactly helpful just yet.

PR #293, opened 2022-07-25
Update overrides documentation

skeep-exogee:
It seems like the statement:

> "Note that the overrides field can only be set at the root of the project."

May, not be a true statement.

After discovering a version conflict in our deployments, we applied the overrides to two packages within our monorepo. This change resolved the version conflict.

I'm sure this statement is here with reason. Do we know why this line is here and why overrides only apply (perhaps a better choice of words is "should") in the root package.json?


skeep-exogee:
Thanks for confirming. I think it must have been a caching issue. I have closed the PR 👍

PR #297, opened 2022-07-27
Add Fig as an autocomplete method

brendanfalk:
[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including pnpm. It looks like this:

image

We think this will help users become more efficient with pnpm CLI, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!


zkochan:
Looks like Fig has some specific code for pnpm: https://github.com/withfig/autocomplete/blob/master/src/pnpm.ts

so I guess it is fine to add it to the website.

PR #301, opened 2022-08-08
Remove personal information

Hedreon:
I know it's like one change from one page but yeah, privacy reasons


zkochan:
What privacy reason? It is random name that I came up with.

PR #302, opened 2022-08-09
Added --ignore-scripts argument to documentation

danielgary:
I noticed --ignore-scripts was missing from the documentation, so I added it here.

### DIFF
--- a/docs/cli/install.md
+++ b/docs/cli/install.md
@@ -131,9 +131,8 @@ Creates a flat `node_modules` structure, similar to that of `npm` or `yarn`.
 * Default: **false**
 * Type: **Boolean**
 
-If true, pnpm does not run scripts specified in package.json files.
-
-Note that commands explicitly intended to run a particular script, such as pnpm run start, pnpm run stop, pnpm run restart, pnpm run test, and pnpm run-script will still run their intended script if ignore-scripts is set, but they will not run any pre- or post-scripts.
+Do not execute any scripts defined in the project `package.json` and its
+dependencies.
 
 ### --filter &lt;package_selector>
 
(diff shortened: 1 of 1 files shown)


docs/cli/install.md
```diff
@@ -126,6 +126,15 @@ after installation is done. To stop the store server, run `pnpm server stop`
...
 **WARNING**: This is highly discouraged.
 
+### --ignore-scripts
+
+* Default: **false**
+* Type: **Boolean**
+
+If true, pnpm does not run scripts specified in package.json files.
+
+Note that commands explicitly intended to run a particular script, such as pnpm run start, pnpm run stop, pnpm run restart, pnpm run test, and pnpm run-script will still run their intended script if ignore-scripts is set, but they will not run any pre- or post-scripts.
```
zkochan, line 136: I don't think ignore-scripts works like this with pnpm. pre/post scripts are never executed by pnpm.

also, ignore scripts is already documented here: https://pnpm.io/npmrc#ignore-scripts

danielgary: This is not accurate.  Scripts are run by pnpm, it does respect the --ignore-scripts argument, and it's not documented on the install documentation.

zkochan: You have copied this description from the npm docs. The second paragraph is not related to the install command, and it doesn't make sense with pnpm because pnpm never runs pre/post scripts by default (https://pnpm.io/cli/run#enable-pre-post-scripts).

danielgary: It does run the `prepare` step however, which is the exact script I looked this up for.  Yes, I copied from `npm` and will update the documentation to make more sense in this context.

PR #307, opened 2022-08-21
docs(how-peers-are-resolved): update structure of node_modules 

nnhjs:
Before 1: 
Screen Shot 2022-08-22 at 22 51 06
After 1:
Screen Shot 2022-08-22 at 22 53 57

Before 2: 
Screen Shot 2022-08-22 at 22 55 02
After 2: 
Screen Shot 2022-08-22 at 22 54 52


zkochan:
I don't understand, why have you added those dirs? Such folders are not created by pnpm.

nnhjs:
```
- foo-parent-1
  - bar@1.0.0
  - baz@1.0.0
  - foo@1.0.0
- foo-parent-2
  - bar@1.0.0
  - baz@1.1.0
  - foo@1.0.0
```
This is the structure of the project so foo@1.0.0 be installed by pnpm so I think the dependency be handled by pnpm.

PR #310, opened 2022-08-29
Update users.json

carlHandy:
Add Kalpa Services


zkochan:
Sorry, the users list is already too long. We're going to reduce it.

PR #316, opened 2022-09-12
Update patch.md

IPWright83:
Expanded slightly that it's the `\user` directory that you need to use with the `patch-commit` command.


zkochan:
I have just changed it recently: https://github.com/pnpm/pnpm/pull/5331

there will be only one directory

IPWright83:
Ah no worries @zkochan . In which case I'll just close this. I found the Youtube video super helpful, but was just lacking a little bit. Hopefully it'll more closely align after your change!

PR #323, opened 2022-09-23
Add information about dependenciesMeta.*.injected

klingenm:
When using Workspaces with a log of dependencies, I have encountered many times that a dependency gets duplicated between the packages if the peer dependencies are satisfied differently. Since packages may have optional peer dependencies this is a valid case that should not require the package to be duplicated.

It took a long time to find a solution to the problem since the solution is only documented under the package.json documentation page.

### DIFF
--- a/docs/workspaces.md
+++ b/docs/workspaces.md
@@ -131,6 +131,14 @@ pnpm cannot guarantee that scripts will be run in topological order if there are
 
 If you see the message `There are cyclic workspace dependencies`, please inspect workspace dependencies declared in `dependencies`, `optionalDependencies` and `devDependencies`.
 
+The default peer dependency resolution can in some cases cause duplicates of peer dependencies between workspace packages. This may lead to unexpected behavior e.g. where local caches within a module are not shared by a package and its library.
+
+One such case is there are transitive peer dependencies that are fulfilled differently (e.g. one installes an optional dependency while the other does not) in two workspace packages.
+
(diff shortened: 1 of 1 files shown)


zkochan:
I think a better solution might be to force single version of that package using overrides.

klingenm:
Problem is that having single versions is not enough.

If you satisfy peer optional peer dependencies differently in two packages, the "real" dependency gets duplicated.

Lets say we have two packages in our workspace called A and B.  They both depend on external package C, that has two optional dependencies D and E.

Now if A only needs D,  B only needs E then C will be duplicated, once with D as a peer dependency and once with E as a peer dependency.

The only way to get around that duplication is to take the sum of all peer dependencies used throughout the workspace and apply them to all who import C.

This example probably would not have any side-effects. But if the peer dependency is in a common lib F, that is imported in A and B we can have a lot of issues if C e.g is @nestjs/core that has a bunch of optional dependencies.

docs/workspaces.md
```diff
@@ -131,6 +131,14 @@ pnpm cannot guarantee that scripts will be run in topological order if there are
 
 If you see the message `There are cyclic workspace dependencies`, please inspect workspace dependencies declared in `dependencies`, `optionalDependencies` and `devDependencies`.
```
zkochan, line 133: Add a title to this section. Maybe "## Peer dependencies in a workspace"

PR #329, opened 2022-09-26
Update workspaces.md to fix mistake in workspaces docs

lancerael:
I think this is a typo? It completely reverses the meaning and is quite confusing,


juanpicado (approved):
Yes, I think you are right, I think the whole sentence might need more visual help, it's a bit hard to  grasp at first glance.

Thanks !

zkochan (approved):

PR #330, opened 2022-09-28
Add PNPM_VERSION info for standalone install script

elliottsj:


docs/installation.md
```diff
@@ -25,6 +25,12 @@ iwr https://get.pnpm.io/install.ps1 -useb | iex
 
 You may use the [pnpm env] command then to install Node.js.
 
+Prior to running the install script, you may optionally set a shell variable `PNPM_VERSION` to install a specific version of pnpm:
+
+```sh
+export PNPM_VERSION=<version>
+```
+
```
zkochan, line 33: I think I would recommend running it with one line

like 

```
pnpm_version=<version> curl -fsSL https://get.pnpm.io/install.sh | sh -
```

elliottsj: Updated, but using the `env` command for compatibility with fish shell, should also work in bash/zsh

zkochan (approved):

PR #333, opened 2022-10-02
docs: fix typesetting of "GitLab"

simon04:
Use the official typesetting of GitLab.

Ref: https://about.gitlab.com/company/


juanpicado (approved):
Thanks

PR #360, opened 2022-12-07
docs: alpine wget instead of curl

moander:
`wget` is included in vanilla alpine (`docker run alpine`)

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -30,7 +30,7 @@ wget -qO- https://get.pnpm.io/install.sh | sh -
 ### On Alpine Linux
 
 ```sh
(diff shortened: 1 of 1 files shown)


juanpicado (approved):
LGTM, thanks for sharing it

zkochan:
curl isn't available in alpine? The old code wasn't working?

moander:
> curl isn't available in alpine? The old code wasn't working?

Try this (should work)
```sh
docker run alpine which wget
```

And this (should fail)
```sh
docker run alpine which curl
```

zkochan (approved):

PR #361, opened 2022-12-09
Add Bazel to the list of pnpm-supporting tools

alexeagle:
Aspect's rules_js replaces Bazel's built-in JS support by leveraging pnpm.

Read https://blog.aspect.dev/rules-js in case you're curious.


alexeagle:
ping @zkochan ?

PR #367, opened 2023-01-03
Grammatical fixes for docs/cli/fetch.md

JoshTheDerf:
Small tweaks to `docs/cli/fetch.md` to tidy up the wording and reduce ambiguity.  Thanks for the great work on pnpm!

### DIFF
--- a/docs/cli/fetch.md
+++ b/docs/cli/fetch.md
@@ -7,13 +7,13 @@ Fetch packages from a lockfile into virtual store, package manifest is ignored.
 
 ## Usage scenario
 
-This command is specifically designed to boost building a docker image.
+This command is specifically designed to improve building a docker image.
 
 You may have read the [official guide] to writing a Dockerfile for a Node.js
-app, if you didn't read it yet, you may want to read it first.
+app, if you haven't read it yet, you may want to read it first.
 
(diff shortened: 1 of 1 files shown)


docs/cli/fetch.md
```diff
@@ -71,8 +71,8 @@ CMD [ "node", "server.js" ]
 As you can see, the Dockerfile has to be updated when you add or remove
 sub-packages.
 
-`pnpm fetch` will solve the above problem perfectly by providing the ability
-to fetch package to virtual store with information only from a lockfile.
+`pnpm fetch` solves the above problem perfectly by providing the ability
+to fetch packages from the virtual store with information only from a lockfile.
```
zkochan, line 75: this changes the meaning of the sentence. Maybe "to the" should be used.

JoshTheDerf: > this changes the meaning of the sentence. Maybe "to the" should be used.

Oops. I've now reworded that a bit to fix the logic.
"Fetch" is an odd word in English, you can only use it to specify where something is coming from, not where it's going to. "He fetched apples from the market", but not "He fetched apples from the market to the car".

zkochan (approved):

PR #372, opened 2023-01-14
make it clearer that `packageExtensions` can only be set at root package

ztl8702:
# Context

I spent quite a bit of time debugging my package.json, until I found this comment https://github.com/pnpm/pnpm/issues/4097#issuecomment-1044454589

I think this caveat should be made clearer in the documentation.

### DIFF
--- a/docs/package_json.md
+++ b/docs/package_json.md
@@ -220,11 +220,15 @@ For example:
 
 ## pnpm.overrides
 
+:::note
+
+If you are using workspace, `overrides` can only be set at the root package.
+
+:::
+
 This field allows you to instruct pnpm to override any dependency in the
 dependency graph. This is useful to enforce all your packages to use a single
 version of a dependency, backport a fix, or replace a dependency with a fork.
-
-Note that the overrides field can only be set at the root of the project.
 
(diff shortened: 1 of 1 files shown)


ztl8702:
There are other sections in this file, e.g. https://pnpm.io/package_json#pnpmneverbuiltdependencies, let me know if I should add the same note to those sections too.

zkochan:
All those fields should be in the root of the workspace. I don't think you should duplicated the info to all fields. Just put it at the top of the file maybe.

Toxaris:
Some documentation that I need to apply these settings in the root package.json would have helped me today with converting a running example to a monorepo setup with pnpm workspaces. Would be great to see this merged!

Toxaris:
On second thought, in addition to documenting this, could `pnpm` have warned me about my ineffective `pnpm` setup in a non-root package.json? Something like:

```
$ pnpm install
WARN: It seems that you are putting pnpm configuration into `frontend/package.json` but pnpm will only read configuration from the root package.json file.
```

zkochan:
adding a warning sounds good.

PR #377, opened 2023-01-17
Update the pnpm to latest version of pnpm

officialrajdeepsingh:
Add guidance on how to update the pnpm latest in documents


zkochan:
No, I don't think we can currently recommend that command. Users need to update pnpm the same way they have installed it. So if they installed pnpm with npm, they need to update it with npm. If they installed it with the standalone script, they should run the standalone script to install the latest version.

officialrajdeepsingh:
Sorry for the late release; I create a separate details page for that topic. When complete, I push it again into the main repository. I hope. It will be helpful for developers.

PR #378, opened 2023-01-18
make it clearer that pnpx doesn't run a binary of a project's dependency

mikib0:
I think it's important to note that pnpx and npx are not completely equivalent, as npx runs the binary of a project's dependency, while pnpx doesn't. This distinction can cause confusion for npm users who may assume that pnpx can be used interchangeably with npx when running binary commands of project dependencies. One specific example of this discrepancy is when the binary command name is different from the package name, as seen in this issue (https://github.com/sverweij/dependency-cruiser/issues/725). To avoid confusion, it's helpful to understand the differences between pnpx and npx when it comes to running binary commands of project dependencies.


zkochan:
These are the docs for v6 in which pnpx worked exactly like npc

PR #385, opened 2023-02-04
chore: for Apline differentiate between x86 and arm

bvanjoi:


zkochan:
I think that separate section for Alpine may removed

The generic script was updated to work with alpine: https://github.com/pnpm/get.pnpm.io/pull/28

zkochan (approved):

PR #390, opened 2023-02-21
Suggest to use corepack to install pnpm in CI

zkochan:
This will only work with node.js 16, so I am not sure if it is too soon or we can update it already.


juanpicado (approved):
LGTM. This is my primary way to use it very handy. :)

PR #391, opened 2023-02-21
docs: fix pnpm version in corepack prepare step

sitek94:
When I followed the docs, I got the following error:

```
Usage Error: Tag not found (latest-7)
```
Shouldn't it be either `pnpm@latest` or `pnpm@7`?

Screenshot 2023-02-21 at 16 19 04

### DIFF
--- a/docs/continuous-integration.md
+++ b/docs/continuous-integration.md
@@ -17,7 +17,7 @@ cache:
     - "~/.pnpm-store"
 before_install:
   - corepack enable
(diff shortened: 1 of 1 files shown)


zkochan:
fixed. Now it should work.

sitek94:
Oh, so it wasn't error in the docs after all :) 

I confirm that it works now:

Screenshot 2023-02-28 at 15 07 18

Thanks @zkochan!

PR #396, opened 2023-03-09
chore: adds rushstack as Usage example

alber70g:


zkochan:
They don't use the pnpm workspace.

PR #401, opened 2023-03-25
add benchmarking of npm linked strategy

asdfjkalsdfla:
Adds benchmarks of the new npm linking approach. (I ran into issues with node_modules so disabled that.)


zkochan:
I don't think it is a good idea to add more stuff to these benchmarks. If we add a separate benchmark for npm linked approach, then we should probably add also Yarn's pnpm linker and pnpm's hoisted linker.

The initial goal of these benchmarks was to compare package managers with their default configurations. For Yarn we did an exception because Plug'n'Play doesn't really fit to these benchmarks.

PR #407, opened 2023-04-13
docs: update npmrc

ntnyq:
Looks like many `.npmrc` settings are not documented.

### DIFF
--- a/docs/npmrc.md
+++ b/docs/npmrc.md
@@ -636,6 +636,28 @@ Regardless of this configuration, installation will always fail if a project
 
 The location of the npm binary that pnpm uses for some actions, like publishing.
 
+### shell-emulator
+
+* Default: **false**
+* Type: **Boolean**
+
+When set to `true`, pnpm will use a JavaScript implementation of a [bash-like shell] to
+execute scripts.
+
+This option simplifies cross-platform scripting. For instance, by default, the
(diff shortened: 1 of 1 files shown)


docs/npmrc.md
```diff
@@ -636,6 +636,28 @@ Regardless of this configuration, installation will always fail if a project
 
 The location of the npm binary that pnpm uses for some actions, like publishing.
 
+### shell-emulator
```
zkochan, line 639: but this is documented on the run command page. So, we will duplicate documentation?

docs/npmrc.md
```diff
@@ -822,6 +844,13 @@ If `foo@1.0.0` is in the workspace and you run `pnpm add foo` in another project
 
 When executing commands recursively in a workspace, execute them on the root workspace project as well.
 
+### ignore-workspace-root-check
```
zkochan, line 847: This one is present on the add command page.

ntnyq:
Since this is the document of `.npmrc` file, all supported settings should be documented here.

For settings like `shell-emulator`, maybe we could add a config links list jumping to this page on run command page.

As for `ignore-workspace-root-check`, there is no document shows it can be configured in `.npmrc` file on add command page, only says add command support `--ignore-workspace-root-check` option.
Still, we could add a config links list on that page.

PR #410, opened 2023-04-19
Highlight caveat for a CLI option

sanjarcode:
Point out the correct way to use the option


zkochan:
Why would you use it as a flag at all? If your project uses pre/post, add this setting to `.npmrc` in the root of the project.

sanjarcode:
You're right. I just want to highlight the combination that doesn't work.

zkochan:
We actually have this already documented here: https://pnpm.io/cli/run#options

PR #415, opened 2023-05-17
docs: Update gitlab and bitbucket CI docs, friendly for developers

motea927:
### Description

Currently the online CI example cannot work, it has two errors,
1. The Node.js must >= 16.14
2. Corepack can not use  `@latest-8 syntax`

So I updated docs, at least it can work, it will friendly for developers

image

### DIFF
--- a/docs/continuous-integration.md
+++ b/docs/continuous-integration.md
@@ -120,10 +120,10 @@ stages:
 
 build:
   stage: build
-  image: node:16.9.0
+  image: node:16.20.0
   before_script:
     - corepack enable
-    - corepack prepare pnpm@latest-8 --activate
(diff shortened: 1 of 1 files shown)


motea927:
I have some test

- ❌ Currently the online CI example (node:16.9.0 + pnpm@latest-8) 
image

- ❌ node:16.9.0 + pnpm@8.5.1
image

- ❌ node:16.14.0 + pnpm@latest-8
image

- ✔️ node:16.20.0 + pnpm@latest-8
image


I know that pnpm@latest need **with Node.js v16.17 or newer**, 

So maybe we need to change doc image node version to newer,

I push new change for it,

Hopefully it will be clear and useful.

PR #419, opened 2023-06-06
fix: update ignore patterns error

await-ovo:
related https://github.com/pnpm/pnpm/issues/6645


zkochan:
So we need to excape "!"? This syntax is pretty ugly.

PR #420, opened 2023-06-06
Update installation.md

GoodNovember:
Hi there, I just added an additional step to the corepack installation instructions that talks about how to enable the PNPM package manager to be used. 

Without this command, it will appear that the installation has done nothing and PNPM is nowhere to be found.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -82,6 +82,12 @@ With Node.js v16.17 or newer, you may install the `latest` version of pnpm by ju
 corepack prepare pnpm@latest --activate
 ```
 
+After installation, you will need to enable `pnpm` in order to use it:
+
+```
+corepack enable pnpm
+```
+
 ## Using npm
 
 We provide two packages of pnpm CLI, `pnpm` and `@pnpm/exe`.


zkochan:
Are you sure?

It seems to work without it:

https://github.com/pnpm/pnpm.github.io/blob/763409f56bc3baf0fb8ce0dfa418f35f10a7d966/.github/workflows/deploy.yml#L27-L30

GoodNovember:
I can only share my experience with PNPM. If my machine is somehow unique, perhaps this addition is not needed, but this is what I needed to do in order to use PNPM on my Windows machine from a PowerShell prompt. Your mileage may vary.

zkochan:
What is your version of node.js?

GoodNovember:
Ah, looks like I was on:

Node: v18.13.0 
PNPM: v8.6.2

Just updated the environment to the latest: 20.3.0 (PNPM env rules, btw)

Perhaps that was the issue?

GoodNovember:
I think I figured out why my machine was different, I was apparently using a hodgepodge of methods to keep Node up to date, in fact several different versions of Node were installed at once and were quietly commingling. Anyway, once that was fixed, things make much more sense. Apologies for the PM.

PR #432, opened 2023-07-11
docs: replace `docker` with `podman`

KSXGitHub:
I mistakenly assumed that all options in docker is the same as podman, turns out it wasn't the case


KSXGitHub:
Then I will have to edit it tomorrow, using the multi staged build suggested by one comment in the thread. But for now, the docs contains a misinformation.

zkochan:
ok, I guess we can have a separate page for podman too.

zkochan (approved):

PR #433, opened 2023-07-11
Document `pnpm install`'s `--no-lockfile` option

22a:
https://pnpm.io/cli/install doesn't currently list this option.

References:
- [pnpm/pnpm/pkg-manager/plugin-commands-installation/src/install.ts#L122-L125](https://github.com/pnpm/pnpm/blob/002f6febe489ffefb004110ee5c83c8c2ff80fd2/pkg-manager/plugin-commands-installation/src/install.ts#L122-L125)
- [pnpm/pnpm/pnpm/test/install/misc.ts#L50-L58](https://github.com/pnpm/pnpm/blob/002f6febe489ffefb004110ee5c83c8c2ff80fd2/pnpm/test/install/misc.ts#L50-L58)

### DIFF
--- a/docs/cli/install.md
+++ b/docs/cli/install.md
@@ -66,10 +66,7 @@ were already installed, regardless of the `NODE_ENV`.
 
 ### --no-lockfile
 
-* Default: **false**
-* Type: **Boolean**
-
-If `true`, pnpm doesn't read or generate a `pnpm-lock.yaml` file.
+Don't read or generate a `pnpm-lock.yaml` file.
 
 ### --lockfile-only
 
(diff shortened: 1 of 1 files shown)


zkochan:
It is documented here: https://pnpm.io/npmrc#lockfile

mansona:
@zkochan I'm not sure but I think there might be a bit of confusion here 🤔 the `--no-lockfile` option that this PR is adding to the documentation is not documented on the url that you're linking to.

I mean I might have missed it but I did a "find in page" for `--no-lockfile` and I didn't see anything

22a:
One final bump on this, if you still feel this is unnecessary I will close it out.

I believe the docs would benefit from this addition - the `install` cli command documents other lockfile related options but not `--no-lockfile` - I opened this PR initially because I needed to read the source code to learn how to achieve this behaviour with `pnpm install`.

Screenshot 2025-02-16 at 13 53 17

zkochan (approved):

PR #448, opened 2023-08-23
docs: add info about using Corepack with `env`

nachoaldamav:


docs/cli/env.md
```diff
@@ -5,8 +5,13 @@ title: "pnpm env <cmd>"
 
 Manages the Node.js environment.
 
+:::danger
+`pnpm env` does not include the binaries for Corepack. If you want to use Corepack to install other package managers, you need to install it separately. See [#4029](https://github.com/pnpm/pnpm/issues/4029)
```
zkochan, line 9: Maybe instead of the link to the issue just include a link to how install corepack separately. Or the script.

nachoaldamav: I added the installation command, looking at the [corepack package.json](https://www.npmjs.com/package/corepack?activeTab=code), it seems that using pnpm to install it is correct (`pnpm add -g corepack`), but maybe it's better to install it with npm?

zkochan (approved):

PR #456, opened 2023-09-26
Add Arch Install

julian-a-avar-c:
I saw Alpine as an option. So I thought it might be useful to add the simple way to install in Arch.


zkochan:
I don't know who owns that package. I don't know if we can trust them.

dcousens (changes requested):
This is not the Arch Linux package for `pnpm`, this is from the untrusted AUR.

If you actually want to add this, the official repository package is https://archlinux.org/packages/extra/any/pnpm/, maintained by @polyzen (https://github.com/polyzen).

And, if you are talking about typical Arch Linux users, you would use `pacman`, not `yay`.

See the [`PKGBUILD`](https://gitlab.archlinux.org/archlinux/packaging/packages/pnpm/-/blob/main/PKGBUILD?ref_type=heads).

PR #463, opened 2023-10-07
docs: document `resolutions` and `pnpm.overrides` merging

stevenpetryk:
Docs change for pnpm/pnpm#7174.


docs/package_json.md
stevenpetryk, line 1: The reason this diff looks so bad is some unterminated JSON above https://github.com/pnpm/pnpm.io/blob/main/docs/package_json.md?plain=1#L514

KaliforniaShell: 👍

PR #466, opened 2023-10-19
Optimize standard SVG logo

imagoiq:
Current standard logo embeds a whole font! Let's optimize it. Export is done from Figma then optimized through SVGOMG


zkochan (approved):

zkochan:
Nice

PR #473, opened 2023-11-02
add tinyhttp to the list of users

talentlessguy:
Hi!

I'd like to propose my project, tinyhttp, to be added to a list of projects using pnpm.

It was added before in #77 but the configs were moved around tinyhttp got lost :(

Since the PR was merged, tinyhttp grew from 250 stars to 2500 stars and pnpm still works great for tinyhttp!


talentlessguy:
thank you very much for merging!

PR #474, opened 2023-11-02
Add --dir option to run.md

Airthee:
We can see that some options are only available in CLI `--help` command but not in the documentation.  
I noticed that the `--dir` is usefull so I wanted to add it.  

Maybe there are other options that should be added ?


zkochan:
That option is available on all commands and is documented on this page: https://pnpm.io/pnpm-cli

Airthee:
Thanks I didn't see it...

PR #495, opened 2024-01-07
Document pre/post scripts not supported

basememara:
Further discussion at https://github.com/pnpm/pnpm/issues/2891


basememara:
pre/post scripts revert will happen in v9 for parity with npm and yarn: https://github.com/pnpm/pnpm/pull/7634

PR #498, opened 2024-01-11
Update add.md

uiolee:


docs/cli/add.md
```diff
@@ -115,7 +115,7 @@ pnpm's default semver range operator.
 Using `--save-peer` will add one or more packages to `peerDependencies` and
 install them as dev dependencies.
 
-### --ignore-workspace-root-check
+### --ignore-workspace-root-check, -w
```
zkochan, line 118: `-w` is a shorthand for `--workspace-root`: https://pnpm.io/pnpm-cli#-w---workspace-root

uiolee: >Adding a new dependency to the root workspace package fails, unless the
>`--ignore-workspace-root-check` or `-w` flag is used.
>For instance, `pnpm add debug -w`.

Sorry, I consider they were the same before

PR #499, opened 2024-01-13
docs: add Material UI to the showcase

oliviertassinari:
I added Material UI in the same sort DESC stars order. The PR in question: https://github.com/mui/material-ui/pull/36287.

The page: https://pnpm.io/workspaces#usage-examples.

Hopefully, this will further help convince developers that pnpm is now a better choice than yarn. I mean, I think that developers who are considering pnpm should see this list, this looks like a very powerful marketing tool to me, maybe to make it easier to discover from the homepage.


zkochan:
Thanks. Yes, we could make it more prominent.

PR #504, opened 2024-01-24
Update Corepack instructions in `installation.md`

aduh95:
The current instructions were a bit outdated, so I'm suggesting a few updates:

- `corepack enable` currently also installs Yarn. It might make sense to restrict it to pnpm.
- Since https://github.com/nodejs/corepack/pull/134, the version folks get when installing Corepack should always be the latest.
- We could talk about package-pinning with `corepack use`.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -67,7 +67,7 @@ If you have installed Node.js with `pnpm env` Corepack won't be installed on you
 :::
 
 ```
-corepack enable
+corepack enable pnpm
 ```
 
 If you installed Node.js using Homebrew, you'll need to install corepack separately:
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -76,18 +76,16 @@ If you installed Node.js using Homebrew, you'll need to install corepack separat
...
 
-This will automatically install pnpm on your system. However, it probably won't be the latest version of pnpm. To upgrade it, check what is the [latest pnpm version](https://github.com/pnpm/pnpm/releases/latest) and run:
+This will automatically install pnpm on your system.
 
-```
-corepack prepare pnpm@<version> --activate
-```
-
-With Node.js v16.17 or newer, you may install the `latest` version of pnpm by just specifying the tag:
+You can pin the version of pnpm used on your project useing the following command:
```
zkochan, line 81:
```suggestion
You can pin the version of pnpm used on your project using the following command:
```

docs/installation.md
```diff
@@ -76,18 +76,16 @@ If you installed Node.js using Homebrew, you'll need to install corepack separat
 brew install corepack
 ```
 
-This will automatically install pnpm on your system. However, it probably won't be the latest version of pnpm. To upgrade it, check what is the [latest pnpm version](https://github.com/pnpm/pnpm/releases/latest) and run:
+This will automatically install pnpm on your system.
 
-```
-corepack prepare pnpm@<version> --activate
```
zkochan, line 82: what if I don't want to use the `packageManager` field just install the latest pnpm version globally?

aduh95: You should expect Corepack to download the latest version on the first use. After that, if you want to manually bump it, you can run `corepack install --global pnpm@latest` (or the old `corepack prepare` command, that still works).

zkochan: So `corepack enable pnpm` installs the latest version? It used to install some hardcoded stable version.

aduh95: The short answer is "yes". The long answer is:
Technically, `corepack enable pnpm` only adds a binary on the PATH; pnpm will be downloaded only when it's first needed (so either when calling `corepack pnpm …` or using the `pnpm` bin created by `corepack enable`), and here are the steps that Corepack takes to determine what version to use:
- if there's a local `package.json` with a `"packageManager"` field, use the version defined there.
- otherwise, check if there's a cached global version, if so, use that.
- otherwise, check if env variable `COREPACK_DEFAULT_TO_LATEST` was set to 0, in which case, use the hard coded version.
- otherwise, use the latest version from the npm registry.
It did use to default to the hardcoded version, but that was a while back.

zkochan (approved):

PR #505, opened 2024-01-27
docs(cli): correct `--latest` to follow merged PR

KSXGitHub:
https://github.com/pnpm/pnpm/pull/7466

### DIFF
--- a/docs/cli/update.md
+++ b/docs/cli/update.md
@@ -60,7 +60,7 @@ pnpm --recursive update typescript@latest
 
 ### --latest, -L
 
-Update the dependencies to their latest stable version as determined by their `latest` tags (potentially upgrading the packages across major versions) as long as the version range specified in `package.json` is older than the `latest` tag (i.e. it will not downgrade prereleases).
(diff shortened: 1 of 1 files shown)


docs/cli/update.md
```diff
@@ -60,7 +60,7 @@ pnpm --recursive update typescript@latest
 
 ### --latest, -L
 
-Ignores the version range specified in `package.json`. Instead, the version specified by the `latest` tag will be used (potentially upgrading the packages across major versions).
+Update the dependencies to their latest stable version as determined by their `latest` tags (potentially upgrading the packages across major versions) as long as the version range specified in `package.json` is older than the `latest` tag (i.e. it will not downgrade prereleases).
```
zkochan, line 63: Instead of "older" I'd say "lower". It won't be necessarily older.

KSXGitHub:
```suggestion
Update the dependencies to their latest stable version as determined by their `latest` tags (potentially upgrading the packages across major versions) as long as the version range specified in `package.json` is lower than the `latest` tag (i.e. it will not downgrade prereleases).
```

PR #508, opened 2024-02-07
docs: completion

KSXGitHub:


KSXGitHub:
For some reason, it doesn't appear in the sidebar.

zkochan:
We can't merge it now. We don't have docs for v9 yet. And this will only ship in v9.

PR #524, opened 2024-04-17
docs: add pwsh in completion

kingyue737:
Maybe we can close https://github.com/pnpm/pnpm/issues/3771 since it is supported in v9.


kingyue737:
duplicated #508

PR #536, opened 2024-05-16
docs: update jetbrains sponsor logo

niklas-wortmann:
We updated our logo for JetBrains IDEs. I hope it is ok that I increased the size, it kinda looked weird next to the others. Hope it wasn't associated with a certain sponsor package or something. If so, I am more than happy to revert that change


zkochan:
This will only look good in the dark theme (which is off by default).

niklas-wortmann:
@zkochan that's a good point, I will test a couple different logo variation. Thanks!

zkochan:
How did you manage to make the logo change its color in the different theme?

niklas-wortmann:
svg has support for the style attribute so I just ship some basic styles with the SVG 
image

PR #539, opened 2024-05-18
chore(env): add install latest version node desc

btea:


btea:
@zkochan Do you think this is necessary?

zkochan:
I don't understand. That page already has an example how to install the latest version of Node.js

btea:
Oh, sorry, I didn't notice it before.

PR #557, opened 2024-07-17
Update test.md

maximvl:
Add documentation regarding custom test options.

I found it in the contribution guide here https://github.com/pnpm/pnpm/blob/726753f05d6512c93dd6f26787a92891ded59533/CONTRIBUTING.md#running-tests

And tested it on my local setup.


zkochan:
I think we should make the `pnpm test` command work with additional args too.

PR #569, opened 2024-08-21
Update Docker and pnpm fetch to be more aligned and link to each other

segevfiner:


segevfiner (commented):
I also added a more complex Dockerfile example that combines all the techniques.

segevfiner:
Also it is likely a user will want a `.dockerignore`/`Dockerfile.dockerignore`, e.g.
```ignore
**/node_modules
**/dist
```

Though there are likely additional patterns he might want. We can add that to the docs as well.

josh-hemphill:
Updates to the docker instructions are needed, but corepack is no longer distributed: https://github.com/pnpm/pnpm.io/issues/719

segevfiner:
If this gets merged I don't mind updating the instructions about corepack. Though we need a decision on whether to recommend to install corepack manually into the images or if we want to add commands to manually install pnpm (Which can be annoying as that won't read the version from the `packageManager` field.

segevfiner:
Still on the hunt for the optimal Dockerfile -.-" I would really wish this got a look over...

karlhorky:
@segevfiner @zkochan I think this can be closed now? Now that there are official Docker images and Corepack was removed from the `Dockerfile`s in in PR #870?

segevfiner:
Might be that some of it is still relevant. But probably less with the new Dockerfile templates and image. It needs to be recreated and can probably be closed in the mean time.

karlhorky:
@segevfiner ok great, go ahead and close when you're ready.

Issue #577, opened 2024-09-16
Improve `pnpm add <git remote URL>` documentation


MichalBryxi:
## The problem

I wanted to install a package from a specific commit on my github fork. And I was a biiit confused around the current section of [Install from git repository](https://pnpm.io/cli/add#install-from-git-repository) documentation.

I went to a deep(ish) dive [through commits](https://github.com/pnpm/pnpm.io/commit/d50d118b337bd8d1de61f215b1e91b6916746664) and [patches/discussions](https://github.com/pnpm/pnpm/issues/4765) to understand how things are and eventually landing on reading [the tests](https://github.com/pnpm/pnpm/blob/main/resolving/git-resolver/test/index.ts).

## The solution

So to prevent future generations needing to go through similar conundrum, I'd propose enhancing said part of documentation to:

- Create sections to make visual/logical distinction of individual parts.
- Spell out all major possible syntaxes.
- Add more explanation to individual parts of the `URL` like `semver:` and `path:`.
- Copy & paste examples directly from [tests](https://github.com/pnpm/pnpm/blob/main/resolving/git-resolver/test/index.ts) so that we're reasonably sure they are correct.
- Explain defaults.
- Explain common cases like monorepo & forks.

PR #578, opened 2024-09-16
docs: Spell out "Install from Git repository" options properly

MichalBryxi:
- Create sections to make visual/logical distinction of individual parts.
- Spell out all major possible syntaxes.
- Add more explanation to individual parts of the `URL` like `semver:` and `path:`.
- Copy & paste examples directly from [tests](https://github.com/pnpm/pnpm/blob/main/resolving/git-resolver/test/index.ts) so that we're reasonably sure they are correct.
- Explain defaults.
- Explain common cases like monorepo & forks.

Fixes: #577

### DIFF
--- a/docs/cli/add.md
+++ b/docs/cli/add.md
@@ -120,7 +120,7 @@ pnpm add RexSkz/test-git-subfolder-fetch#path:/packages/simple-react-app
 
 #### Full URL
 
-If you want to be more explicit or are using alternative git hosting, you might want to spell out full git URL:
+If you want to be more explicit or are using alternative Git hosting, you might want to spell out full Git URL:
 
 ```
 # git+ssh
--- a/docs/cli/add.md
+++ b/docs/cli/add.md
@@ -84,13 +84,34 @@ Installs the package from the hosted Git provider, cloning it with Git.
 
 You may install packages from Git by:
 
-* latest commit from default branch: `pnpm add kevva/is-positive`
-* git commit hash: `pnpm add kevva/is-positive#97edff6f525f192a3f83cea1944765f769ae2678`
-* git branch: `pnpm add kevva/is-positive#master`
-* git branch relative to refs: `pnpm add zkochan/is-negative#heads/canary`
-* git tag: `pnpm add zkochan/is-negative#2.0.1`
-* v-prefixed git tag: `pnpm add andreineculau/npm-publish-git#v0.0.7`
-* version range: `pnpm add kevva/is-positive#semver:^2.0.0`
+* Latest commit from default branch:
+```
+pnpm add kevva/is-positive
+```
+* Git commit hash:
+```
+pnpm add kevva/is-positive#97edff6f525f192a3f83cea1944765f769ae2678
+```
+* Git branch:
+```
+pnpm add kevva/is-positive#master
+```
+* Git branch relative to refs:
+```
+pnpm add zkochan/is-negative#heads/canary
+```
+* Git tag:
+```
+pnpm add zkochan/is-negative#2.0.1
+```
+* V-prefixed Git tag:
+```
+pnpm add andreineculau/npm-publish-git#v0.0.7
+```
+* Version range:
+```
+pnpm add kevva/is-positive#semver:^2.0.0
+```
 
 ### Semver
 
--- a/docs/cli/add.md
+++ b/docs/cli/add.md
@@ -117,17 +117,20 @@ pnpm add kevva/is-positive#semver:^2.0.0
 
 You can specify version (range) to install using the `semver:` parameter. For example:
 
+* Strict semver:
 ```
-# strict semver
 pnpm add zkochan/is-negative#semver:1.0.0
-
-# v-prefixed strict semver 
+```
+* V-prefixed strict semver:
+```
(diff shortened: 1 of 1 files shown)


MichalBryxi:
Happy to talk about any and all choices and alter things if necessary, but I firmly believe that this amount of veboseness greatly reduces user confusion as they don't have to guess what is possible / why / how.

docs/cli/add.md
```diff
@@ -81,26 +81,87 @@ pnpm add <git remote url>
...
+```
+
+#### Providers shorthand
+
+You can use a protocol shorthand `[provier]:` for certain Git providers:
+
+```
+pnpm add github:zkochan/is-negative
+pnpm add bitbucket:pnpmjs/git-resolver
+pnpm add gitlab:pnpm/git-resolver
```
MichalBryxi, line 152: Is there more? I found only those in tests.

docs/cli/add.md
```diff
@@ -81,26 +81,87 @@ pnpm add <git remote url>
...
-`pnpm add github:user/repo`
 
-You may install from Git by:
+You may install packages from Git by:
 
-* latest commit from master: `pnpm add kevva/is-positive`
-* commit: `pnpm add kevva/is-positive#97edff6f525f192a3f83cea1944765f769ae2678`
-* branch: `pnpm add kevva/is-positive#master`
-* version range: `pnpm add kevva/is-positive#semver:^2.0.0`
+```
```
MichalBryxi, line 87: Personally I prefer these triple-fenced examples, as they are way easier to target with a mouse for copy&pasting as you're constrained by the newlines. As opposed to the bullet point format above, which allows for copying the non-code prefix as well.

docs/cli/add.md
```diff
@@ -81,26 +81,87 @@ pnpm add <git remote url>
...
+pnpm add zkochan/is-negative#2.0.1
+
+# v-prefixed git tag
+pnpm add andreineculau/npm-publish-git#v0.0.
+```
+
+### Semver
 
-You may also install just a subdirectory from a Git-hosted monorepo. For instance:
+You can specify version (range) to install using the `semver:` parameter. For example:
```
MichalBryxi, line 109: I consider the `semver:` and `path:` (below) being explicitly spelled out very important. It makes sure that the user knows it's a keyword and not just a random thing in given repo.

docs/cli/add.md
```diff
@@ -81,26 +81,87 @@ pnpm add <git remote url>
...
 
-You may install from Git by:
+You may install packages from Git by:
 
-* latest commit from master: `pnpm add kevva/is-positive`
-* commit: `pnpm add kevva/is-positive#97edff6f525f192a3f83cea1944765f769ae2678`
-* branch: `pnpm add kevva/is-positive#master`
-* version range: `pnpm add kevva/is-positive#semver:^2.0.0`
+```
+# latest commit from default branch
```
MichalBryxi, line 88: As far as I can tell `pnpm` does not look for `master` per-se, it looks for (in GitHub terminology) [default branch](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/changing-the-default-branch), hence the naming change here.

docs/cli/add.md
zkochan: you can have both. code sections inside bullet points. I don't think there's a reason to move the descriptions into the code blocks.

docs/cli/add.md
```diff
@@ -81,26 +81,87 @@ pnpm add <git remote url>
...
+
+You may also install just a subdirectory from a Git-hosted monorepo using the `path:` parameter. For instance:
+
 ```
+pnpm add RexSkz/test-git-subfolder-fetch#path:/packages/simple-react-app
+```
+
+#### Full URL
+
+If you want to be more explicit or are using alternative git hosting, you might want to spell out full git URL:
```
zkochan, line 135: Git should be capitilized
```suggestion
If you want to be more explicit or are using alternative Git hosting, you might want to spell out full Git URL:
```

docs/cli/add.md
MichalBryxi: No hard feelings, I can revert it, but to show why I'm against copy&paste examples in inline form: Here you can see how precise I have to be to select _exactly_ what I intended:



Compare it to the code-fence version that has the _full line_ for mousing around and thus it's way harder to select unwanted things:

MichalBryxi: I pushed the revert to adhere to the ways it was. Let me know if it works for you now.

zkochan: But you could do it like this:

```
* Strict semver:
  
  `` `
  pnpm add zkochan/is-negative#semver:1.0.0
  `` `
```

MichalBryxi: Oooh, yeah why not. Pushed given change for both blocks. Also I used "Sentence case:" for the bullet points. Let me know if this works for you.

MichalBryxi:
Preview looks reasonable (to me).

zkochan:
merged https://github.com/pnpm/pnpm.io/commit/2be27c08dc0b44d9412e802a5cf40f55ee1ed494

PR #583, opened 2024-09-30
Update exec.md

hcpty:
For example, if you changed to a subdirectory called "foo/" which has a "bar.css" under it, and then you execute "pnpm exec postcss bar.css", it won't work, postcss won't find bar.css, but if you execute "pnpm exec postcss foo/bar.css", postcss will find bar.css as foo/bar.css, it is weird.

### DIFF
--- a/docs/cli/exec.md
+++ b/docs/cli/exec.md
@@ -3,7 +3,8 @@ id: exec
 title: pnpm exec
 ---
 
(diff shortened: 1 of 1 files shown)


zkochan:
I am not sure what would be the expected behaviour. How does it work with npm/yarn?

PR #586, opened 2024-10-17
Document missing pnpmfile hooks

Silic0nS0ldier:
Adds documentation for;
* `filterLog`
* `fetchers`
* `importPackage`
* `preResolution`

Local/global pnpmfiles and how they interact is now also documented. For example `readPackage` is supported by local and global pnpmfiles, and both run with global going first so `(global then local)`.

### DIFF
--- a/docs/pnpmfile.md
+++ b/docs/pnpmfile.md
@@ -14,12 +14,16 @@ lockfile. For instance, in a [workspace](workspaces.md) with a shared lockfile,
 
 ### TL;DR
 
-| Hook Function                                         | Process                                                    | Uses                                               |
-|-------------------------------------------------------|------------------------------------------------------------|----------------------------------------------------|
-| `hooks.readPackage(pkg, context): pkg`                | Called after pnpm parses the dependency's package manifest | Allows you to mutate a dependency's `package.json` |
-| `hooks.afterAllResolved(lockfile, context): lockfile` | Called after the dependencies have been resolved.          | Allows you to mutate the lockfile.                 |
+| Hook Function                                         | Process                                                     | Uses                                                      |
+|-------------------------------------------------------|-------------------------------------------------------------|-----------------------------------------------------------|
+| `hooks.afterAllResolved(lockfile, context): lockfile` | Called after the dependencies have been resolved.           | Mutate the lockfile.                                      |
+| `hooks.readPackage(pkg, context): pkg`                | Called after pnpm parses the dependency's package manifest. | Mutate a dependency's `package.json`                      |
+| `hooks.filterLog(log): boolean`                       | Called before a log is printed.                             | Filter logs by level and/or message content.              |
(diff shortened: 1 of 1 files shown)


zkochan:
I am not sure we want people to know about these. All these hooks except `filterLog` were added for easier integration with [stackblitz](https://stackblitz.com/). Would anyone else want to use these low level hooks?

Silic0nS0ldier:
> Would anyone else want to use these low level hooks?

The use case I have in mind sourcing package data directly from the [`npm_package`](https://docs.aspect.build/rulesets/aspect_rules_js/docs/npm_package) rule (Bazel ruleset), such that content is identical across Bazel builds and external tools that use pnpm's `node_modules` (including IDEs).

There are alternatives.

1. Using Bazel's `node_modules` instance instead, a crude implementation of which exists [here](https://github.com/Silic0nS0ldier/vscode-git-monolithic-extension/blob/20ce69f5b04be636b9d8ca5a2d0f2f3d69324c5e/reveal-node-modules.sh), however.
   * This breaks the `pnpm patch` workflow which relies on pnpm's `node_modules` existing.
     Related: https://github.com/pnpm/pnpm/issues/8257
   * pnpm is able to do in-place mutation of `node_modules` which is faster than than copying over.
     Symlinking would close this gap though (why didn't I think to do that before...)
2. Write generated sources into version control (what we are doing at Canva right now).
   * Prevents in-place modification of files. This pattern is unusual in the NodeJS ecosystem, but in the context of our build system such transforms are useful it allows input sources to be normalised (stripping comments, etc) to avoid unnecessary rebuilds, however.
   * Packages in pnpm's `node_modules` include more files than intended (e.g. `.ts` sources).
   * Need to confirm generated sources are up-to-date on CI.
   * Devs need to run an extra step whenever such packages are modified (a common complaint, especially when testing changes to a workflow like ESLint which operates outside of our build system).
   * More noise to sift through in code review (even with Github hiding the diff of these generated files).
3. Redirect `workspace:*` references to tarballs/directories built by Bazel (e.g. by mutating the lockfile). This alternative came about from ideation and has not been implemented. The cons of this haven't had much thought at this stage, off the top of my head it could affect dependency resolution (workspace packages are already by default something of a special case unless they are injected).

The use case I give likely runs into the same dependency resolution issue as mentioned in (3). The issue is not insurmountable (a great deal of flexibility is afforded across all the hooks) but would need to be carefully thought out.

Silic0nS0ldier:
@zkochan I noticed pnpm v10.10 allows `preResolution`, `importPackage` and `fetchers` hooks in the local `.pnpmfile.cjs`. Has their scope (stackblitz integration) expanded?

zkochan:
A few other companies are planning to use the hooks. I added short docs about them: https://github.com/pnpm/pnpm.io/commit/4e852ad5052cb3217560d72643522ce2bbea139c

also, it is now easier to ship hooks via dependencies, using [configDependencies](https://pnpm.io/config-dependencies).

PR #588, opened 2024-10-19
Patch `react-dom` and suppress NUL in non-English docs

tats-u:
Some non-English pages in pnpm.io has NUL (U+0000). 

https://github.com/facebook/docusaurus/issues/9985#issuecomment-2139649267

This is a bug in React 18.1.0–18.3.1. The fix (https://github.com/facebook/react/pull/26228) is available only in 19.x or later. You should patch `react-dom` until you upgrade React to 19.x.

https://github.com/facebook/react/issues/31134#issuecomment-2416257221
https://github.com/tats-u/react-dom-no-nul
https://github.com/tats-u/react-dom-no-nul/commit/850327cbc752bf95e94114699861d142741f4181

↓patched part in `*.production.min.js`:

diff

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,5 +40,10 @@
     "@types/node": "^20.12.2",
     "shx": "^0.3.4",
     "ts-node": "10.9.2"
+  },
+  "pnpm": {
+    "patchedDependencies": {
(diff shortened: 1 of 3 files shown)


tats-u:
Superseded by #698 because latest Docusaurus is now compatible with React 19.

PR #596, opened 2024-11-21
Add an example to `docs/pnpm-workspace_yaml.md` that doesn't use a wildcard

AlexErrant:
While submitting [another issue](https://github.com/vercel/turborepo/issues/9490), I realized that it'd be nice to document how to specify a package directory without a wildcard. It's pretty basic/obvious, but I think it'd be nice to state it explicitly.

### DIFF
--- a/docs/pnpm-workspace_yaml.md
+++ b/docs/pnpm-workspace_yaml.md
@@ -11,6 +11,8 @@ For example:
 
 ```yaml title="pnpm-workspace.yaml"
 packages:
+  # specify a package dir
+  - 'my-app/'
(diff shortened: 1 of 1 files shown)


docs/pnpm-workspace_yaml.md
```diff
@@ -11,6 +11,8 @@ For example:
 
 ```yaml title="pnpm-workspace.yaml"
 packages:
+  # specify a package dir
+  - 'my-app/'
```
zkochan, line 15: I don't think the trailing slash is required.

PR #623, opened 2025-01-10
chore: updated translation progress

Mister-Hope:
🎉 Now zh-CN is 100% complete


zkochan:
We have a job that automatically updates the translation stats. No need to do it manually.

Mister-Hope:
Fine, then plz update the website.

Mister-Hope:
I wanna to check the real website for formatting issues

PR #624, opened 2025-01-12
feat: add ignoredBuiltDependencies

btea:
sync https://github.com/pnpm/pnpm/pull/8958

### DIFF
--- a/docs/package_json.md
+++ b/docs/package_json.md
@@ -508,7 +508,9 @@ The JSON file itself should contain an array of package names:
 
 ## pnpm.ignoredBuiltDependencies
 
-When a package is in the list, pnpm will not print an info message about that package not being built.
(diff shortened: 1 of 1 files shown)


zkochan:
I'll merge it when we publish the feature.

docs/package_json.md
```diff
@@ -506,6 +506,20 @@ The JSON file itself should contain an array of package names:
 ]
 ```
 
+## pnpm.ignoredBuiltDependencies
+
+When a package is in the list, pnpm will not print an info message about that package not being built.
```
zkochan, line 511:
```suggestion
## pnpm.ignoredBuiltDependencies

Added in: v10.1.0

A list of package names that should not be built during installation.
```

PR #644, opened 2025-02-21
docs: add note about PNPM catalog support in version 9.5.0 and update…

anasmohammed361:
… pnpmfileChecksum


zkochan:
These are the docs for pnpm v10. All v10 versions have the catalogs support.

The v9 docs do specify in which version the feature was shipped: https://pnpm.io/9.x/catalogs

image

PR #658, opened 2025-04-07
docs(settings): add `reporter` as a valid CLI and config file option

derekkramer:
Updated `settings.md` to reflect the addition of `reporter` as a valid CLI and config file option. Default setting and possible settings reflect help command output from `cli/common-cli-options-help/src/index.ts` within [pnpm/pnpm](https://github.com/derekkramer/pnpm/blob/main/cli/common-cli-options-help/src/index.ts)


zkochan:
Looks like the info in help is outdated. These are the right docs: https://pnpm.io/cli/install#--reportername

PR #665, opened 2025-05-14
docs(run): add --sequential

PeterCardenas:
The --help option for `pnpm run` gives a `--sequential` option but it is not shown in the documentation, so we do that in this PR.


docs/cli/run.md
```diff
@@ -109,6 +109,12 @@ immediately in all matching packages with prefixed streaming output. This is the
 preferred flag for long-running processes over many packages, for instance, a
 lengthy build process.
 
+### --sequential
+
+Alias for `--workspace-concurrency=1`. Run scripts one after another in a blocking
+manner rather than concurrently. Note that there are no guarantees that the scripts
+will run in topological sort order.
```
MattIPv4, line 116: I noted this in #672, which is a dupe of this, but it'd be good to clarify here what the expected behaviour is in terms of order of execution when using this with `pnpm run "/regex/"`. My understanding is that in that situation, it'll use the order that the scripts are defined in `package.json`.

PeterCardenas: that's fine but i don't think it should block merging this pr. the reason why there's a dupe im assuming is because we both were trying to find this in the docs but needed to look at the help to find it. the pre work for making something more substantive here would be to have tests in the pnpm repo asserting a particular behavior, because i don't think just looking at the implementation is good enough

MattIPv4: I would probably disagree, I think it'd be better to properly figure out what should be documented first, rather than landing partial documentation that doesn't even acknowledge all the use-cases for it (nothing about this wording implies it can be used w/ `pnpm run "/regex/"`, and in some ways even suggests its not for that due to the reference to topological sorting for `pnpm -r`).

PR #691, opened 2025-07-21
docs: added alternative to only allow pnpm

karthikappiah:
Hello team!

I propose a very **_non-invasive update_** to the documentation—a **_native solution_** to a bug that has spawned numerous issues on GitHub.

As described in this [issue](https://github.com/pnpm/only-allow/issues/33#issue-3246739364) (and many more open and closed issues), the package `only-allow` cannot reliably prevent NPM commands from running (even as a preinstall script).

Therefore, I added an alternative—a _native_ solution that simply specifies PNPM as `packageManager` in `package.json` and sets `engine-strict` as `true` in `.npmrc`. This prevents commands from anything other than what's specified as the package manager.

I also documented the option to install only-allow as a dev dependency for offline and CI/CD use cases.

Thank you,
Karthik Appiah

PR #692, opened 2025-07-21
docs(only-allow-pnpm): added option to enforce pnpm with package.json and .npmrc

ZeLinguist:
# TLDR

I added a non-invasive update to a single page of the docs. Currently, the docs only tell you to add `npx only-allow pnpm` to the `preinstall` script inside `package.json`, the option—to set the `.npmrc` configuration variable `engine-strict` as true and specify PNPM as the enforced package manager in the `package.json`—is not in the docs, which is why I added this option to the docs and created a pull request.

## How to Enforce PNPM as the Only Package Manager

For easier reference, here is one of many GitHub threads that describe the same [issue](https://github.com/pnpm/only-allow/issues/33#issuecomment-3094948348):

As per the NPM [docs](https://docs.npmjs.com/cli/v11/configuring-npm/package-json), to prevent the use of an unspecified package manager, from Node v16.9.0 (backported to v14.19.0) onwards, you should:

1. Create a file, if it doesn't already exist, named `.npmrc` at the root of your project.
2. Set the following configuration variable in your `.npmrc` to `true`:

```
engine-strict=true
```

3. Specify the following fields in your `package.json`:

```
{
  "devEngines": {
    "runtime": {
      "name": "node",
      "onFail": "error"
    },
    "packageManager": {
      "name": "pnpm",
      "version": "10.13.1",
      "onFail": "error"
    }
  },
  "engines": {
    "node": ">=18.18.0",
    "pnpm": ">=10.0.0"
  },
}
```

- Now, when you run `npm i` or `npm i -D`, these commands return this error (before the preinstall script can run):

```
username@hostname nodejs-project % npm i -D package
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid engine "packageManager"
npm error EBADDEVENGINES Invalid name "pnpm" does not match "npm" for "packageManager"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'npm', version: '10.9.2' },
npm error EBADDEVENGINES   required: { name: 'pnpm', onFail: 'error' }
npm error EBADDEVENGINES }
npm error A complete log of this run can be found in: /Users/username/.npm/_logs/2021-08-21T00_00_00_000Z-debug-0.log
```

### DIFF
--- a/docs/only-allow-pnpm.md
+++ b/docs/only-allow-pnpm.md
@@ -1,21 +1,51 @@
 ---
 id: only-allow-pnpm
-title: Only allow pnpm
+title: Only Allow PNPM
 ---
 
-When you use pnpm on a project, you don't want others to accidentally run
-`npm install` or `yarn`. To prevent devs from using other package managers,
-you can add the following `preinstall` script to your `package.json`:
+When many developers are working on the same project together, you need a failsafe in case someone accidentally runs commands with another package manager (like NPM, Yarn, Bun).
 
-```json
+To prevent dependency management conflicts between package managers:
+
+1. Create a file, if it doesn't already exist, named `.npmrc` at the root of your project.
+2. Write the following content into your `.npmrc`:
+
+```
+engine-strict=true
+```
+
+3. Write the following content into your `package.json`:
+
+```
 {
-	"scripts": {
-		"preinstall": "npx only-allow pnpm"
-	}
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "onFail": "error"
+    },
+    "packageManager": {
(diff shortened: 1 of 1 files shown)


karthikappiah:
I created a different [pull request](https://github.com/pnpm/pnpm.io/pull/691) previously with suboptimal conventions, so I renamed my branch more descriptively and more concisely explained what I changed and why.

We should consider a pull request template (like a form someone must fill out), similar to how the NextAuth documentation handles pull requests. I believe this will make merging pull requests more productive.

PR #705, opened 2025-09-22
fix example cmd and add pattern in aliases.md

maerzhase:
The docs for aliases display a wrong command. This PR fixes that and also adds a general explanation of the pattern to be used. 

- related to https://github.com/pnpm/pnpm/issues/5645

### DIFF
--- a/docs/aliases.md
+++ b/docs/aliases.md
@@ -17,11 +17,21 @@ Publish a new package called `awesome-lodash` and install it using `lodash` as
 its alias:
 
 ```
-pnpm add lodash@npm:awesome-lodash
+pnpm add awesome-lodash@npm:lodash
 ```
 
 No changes in code are needed. All the requires of `lodash` will now resolve to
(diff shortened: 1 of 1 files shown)


docs/aliases.md
```diff
@@ -17,12 +17,22 @@ Publish a new package called `awesome-lodash` and install it using `lodash` as
 its alias:
 
 ```
-pnpm add lodash@npm:awesome-lodash
+pnpm add awesome-lodash@npm:lodash
```
zkochan, line 20: This was correct. Why have you changed it? Read the paragraph above it.

maerzhase: ah ok jeah now I understand the example. closing the PR then. 

from where I was coming from the example was a bit misleading. i actually wanted to use an explicit version of a package through an alias. so my usecase was e.g.

```
pnpm lodash-old@npm:lodash@3.3.0
```

cheers and thanks for pnpm ;)

PR #706, opened 2025-09-23
Document default of `--location` for `pnpm config`

karlhorky:
Document the default of the `--location` flag for `pnpm config` commands

It appears that the default of the `--location` setting for `pnpm config` is `global`:

```bash
$ pnpm config set x true
$ pnpm config get x
true
$ cat "$HOME/Library/Preferences/pnpm/rc"
x=true
```


karlhorky:
Thanks for the review and merge!

PR #771, opened 2026-04-02
docs: Upgrade Node.js version in Dockerfile

miso-kyoungminkim:
This PR updates the Dockerfile based on the official pnpm guide.
The base image is upgraded from Node.js 20 to Node.js 24 as Node.js 20 is nearing end of LTS.
It also adds an explicit Corepack installation to align with changes in newer Node.js versions.
These updates ensure pnpm works reliably in the container environment.

## Summary by CodeRabbit

* **Documentation**
  * Updated Docker examples to reference Node.js 24, ensuring compatibility with the latest Node.js version in provided configurations.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -34,9 +34,10 @@ dist
 ```
 
 ```dockerfile title="Dockerfile"
-FROM node:20-slim AS base
+FROM node:24-slim AS base
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
+RUN npm install --global corepack@latest
 RUN corepack enable
 COPY . /app
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Node.js 20 reached [end-of-life](https://github.com/nodejs/release#readme) on 2026-04-30, so it definitely makes sense to update the examples to Node.js 24.

MikeMcC399:
Corepack installation isn't necessary for Node.js 24 however, since this version still bundles Corepack.

miso-kyoungminkim:
> Corepack installation isn't necessary for Node.js 24 however, since this version still bundles Corepack.

Oh, you’re right. My bad — fixed it. Thanks!

zkochan (approved):

PR #774, opened 2026-04-11
docs(settings): Drop react-native note from `hoisted` docs

kitten:
Proposal to drop the "A React Native project will most probably only work if you use a hoisted `node_modules`."

There's a bit of nuance to this, because not all React Native projects are the same or set up the same. That said, one recommendation is to use a framework, e.g. Expo: https://reactnative.dev/docs/environment-setup

While I can't vouch for all React Native dependencies and setups to be isolated-safe, at Expo, we've actively worked on this and now support and guarantee that isolated installations work. In fact, we're looking to move this to be the default recommendation for pnpm gradually. We've also in the meantime have started switching the `expo/expo` monorepo to use pnpm in isolated mode as well.

There's kind of a longer story to this, [some of which I've written about in my blog post](https://kitten.sh/blog/autolinkings-broken-promise), but as a direct consequence, I'd love to see:
- the generic "React Native project" example removed from this section of the docs
- and; if it does need to be preserved, it shouldn't make generic statements that may change or are affected by non-Expo projects or make implications that users might generalise to apply to Expo projects

Cheers!

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -447,7 +447,7 @@ Defines what linker should be used for installing Node packages.
 
 * **isolated** - dependencies are symlinked from a virtual store at `node_modules/.pnpm`.
 * **hoisted** - a flat `node_modules` without symlinks is created. Same as the `node_modules` created by npm or Yarn Classic. One of Yarn's libraries is used for hoisting, when this setting is used. Legitimate reasons to use this setting:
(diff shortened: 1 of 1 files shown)


byCedric:
To also provide some historical context:
- https://github.com/pnpm/pnpm/issues/4286
- https://github.com/bycedric/expo-monorepo-example

PR #778, opened 2026-04-27
Update font-family in customTheme.css

314systems:
Switch base font stack to modern system UI fonts for better consistency and maintainability.

### DIFF
--- a/src/css/customTheme.css
+++ b/src/css/customTheme.css
@@ -6,7 +6,7 @@
   --ifm-color-primary-dark: #F0850A;
   --ifm-color-primary-darker: #E37D09;
   --ifm-color-primary-darkest: #BB6708;
-  --ifm-font-family-base: -apple-system,Segoe UI,Roboto,Ubuntu,Cantarell,Noto Sans,sans-serif,BlinkMacSystemFont,"Segoe UI",Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol";
+  --ifm-font-family-base: ui-sans-serif,system-ui,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji";
 }
 
 .header-github-link::before {


314systems:
Using `system-ui (ui-sans-serif)` ensures that the most appropriate font is used (even if users are using different locales).

Before



After

tats-u:
This reverts https://github.com/pnpm/pnpm.io/pull/446.

https://github.com/withastro/starlight/pull/3729
https://github.com/facebook/docusaurus/issues/11727

You shouldn't use `sysetm-ui` other than WebView in desktop apps or widgets.

tats-u:
> Before:
> 

Do you know what font is applied here? It seems a CJK font, which treats Cyrillic letters as double-width, not a font optimized for Latin and Cyrillic scripts.

314systems:
Hiragino Sans (Hiragino Kaku Gothic ProN)

(In Japanese fonts, some Greek and Cyrillic characters are implemented with full-width metrics for backward compatibility with the JIS/Shift_JIS era and for consistency with Japanese fixed-width typesetting. This is especially common in fonts designed for Japanese documents, monospaced layouts, or compatibility-focused use cases. On the other hand, many modern multilingual fonts use proportional widths for Greek and Cyrillic characters, just like Latin text. For example: https://github.com/IBM/plex/issues/428)

tats-u:
> Hiragino Sans (Hiragino Kaku Gothic ProN)

That's a really bad news. It's a highest quality Japanese font but not compatible with non-Japanese languages at all. As you suspect, it should not be applied to Russian (and all European languages) pages because not only Cyrillic but Latin letters in it are not optimal for their native languages.

- Is Hiragino Sans derived from `-apple-system` or `sans-serif`?
- If `sans-serif`, what font is assigned to sans-serif in your browser in Russian? Hiragino Sans or another one?
- Does https://starlight.astro.build/ru/getting-started/ looks great for your environment?

314systems:
> Is Hiragino Sans derived from -apple-system or sans-serif?
`sans-serif`

> If sans-serif, what font is assigned to sans-serif in your browser in Russian? Hiragino Sans or another one?
`Hiragino Kaku Gothic ProN`

> Does <https://starlight.astro.build/ru/getting-started/> looks great for your environment?
Yes. However, that is due to system-ui.

スクリーンショット 2026-05-11 9 05 00
スクリーンショット 2026-05-11 9 06 14
スクリーンショット 2026-05-11 9 06 27
スクリーンショット 2026-05-11 9 10 36

tats-u:
> `sans-serif`
>
> `Hiragino Kaku Gothic ProN`

I see. Chromim browsers shamefully can't configure fonts for non-native languages. You need to use instead e.g.: https://chromewebstore.google.com/detail/advanced-font-settings/caclkomlalccbpcdllchkeecicepbmbm We may need to report this issue in https://issues.chromium.org/ or https://github.com/w3c/csswg-drafts.

> Yes. However, that is due to system-ui.

No, it's thanks to either of -apple-system and BlinkMacSystemFont. `-system-ui` isn't used there.

Probably your screenshots changed the font stack in F12 menu.

> スクリーンショット 2026-05-11 9 05 00

Does this mean you disabled system-ui there by quotation?

> スクリーンショット 2026-05-11 9 06 27

What if you add `BlinkMacSystemFont` or replace `-apple-system` with `BlinkMacSystemFont`?

Anyway we Japanese shouldn't suggest using `system-ui` in general websites. Yu Gothic UI mapped in Japanese Windows is too condensed and hard to be read compared to Noto Sans CJK (Source Han Sans) and Meiryo. You should read to know why kana letters in UI fonts like Yu Gothic UI, Meiryo UI, and MS UI Gothic are so condensed: https://pc.watch.impress.co.jp/docs/column/config/748846.html https://www.nishishi.com/Microsoft/face.html

zkochan:
can't you change the font just for the japanese localization?

tats-u:
You can do it in Firefox, but you can't in Chrome and Edge without such extensions. You can designate only a single font per style. I don't have Mac but I heard Safari can't configure fonts in GUI (You have to tweak global  stylesheet directly) .

tats-u:
Disclaimer: I don't know much about Chrome's behavior because I'm just a Firefox fanboy.

314systems:
It seems that using `BlinkMacSystemFont` causes the San Francisco font to be used (although its behavior appears to differ slightly from Safari). Since San Francisco includes Latin, Greek, and Cyrillic glyphs, it is applied before falling back to Hiragino Sans.

I also tried the following font stack, and it seems to work well as a workaround:

```css
font-family: ui-sans-serif, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", "Noto Sans", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"
```

(Also, I believe this issue is caused by Chromium. At least on macOS, I could not reproduce the same problem in Safari or Firefox.)

314systems:
Also, it appears that in Chromium on Mac, `BlinkMacSystemFont` is converted to `system-ui`.

BlinkMacSystemFont

zkochan:
but the japanese localization is at an entirely different URL. So you should be able to just have different styles for it.

tats-u:
> but the japanese localization is at an entirely different URL.

And a different `lang` attribute value.

> So you should be able to just have different styles for it.

Ideally, you wouldn't need to do this, and could use language-independent specifications like sans-serif to enable each language to use the ideal font stack.

> It seems that using `BlinkMacSystemFont` causes the San Francisco font to be used 

Glad to hear that. I was relieved to find that this designation is still valid today.

I prepared a test case that causes issues with Japanese fonts by using Greek and Cyrillic characters:

https://codepen.io/tats-u/pen/EaNydvy

image
↑This is taken in Edge in Windows Sandbox.

| Language | Font |
| -- |  -- |
| Japanese, English | Noto Sans JP |
| Chinese (Simplified) | Microsoft YaHei |
| Korean | Malgun Gothic |
| Greek, Ukrainian, Russian | Arial |

In my non-sandboxed Chrome, the Chinese font is changed to Noto Sans SC and Korean one is changed to Noto Sans CJK KR because I have installed Noto Sans SC, Noto Sans TC, and Noto Sans CJK. In a non-Japanese Windows environment, Noto Sans JP will be changed to Yu Gothic (`\` will be displayed as `¥` and the weight will get narrower).

In the macOS version of Chromium, at least in Japanese environments, the font fallback for Cyrillic languages appears to be not working properly. I believe it should be reported in https://issues.chromium.org/.

> Since San Francisco includes Latin, Greek, and Cyrillic glyphs, it is applied before falling back to Hiragino Sans.

If San Francisco is used for Japanese text, Hiragino Kaku Gothic is slightly adjusted (e.g. font size gets slightly smaller by 1pt).

https://terkel.jp/archives/2018/07/san-francisco-and-japanese-font/ (Japanese)

PR #782, opened 2026-04-29
fix migration command

aklkv:


zkochan:
pnx should work too. It was added as a short alias into pnpm v11

aklkv:
but if you are migrating to v11 from v10 shorthand doesn't exist yet, am i wrong?

PR #789, opened 2026-05-08
Fix link in release notes for v11.0 migration guide

maurer2:
Hello,

The [v11 release notes](https://pnpm.io/blog/releases/11.0)￼ currently link to an incorrect migration guide page. The correct migration guide is available here: https://pnpm.io/migration.

Cheers

## Summary by CodeRabbit

* **Documentation**
  * Updated the “Migrating from v10 to v11” links in the v11.0 release notes to point to the canonical migration guide instead of a versioned path, ensuring users are directed to the current, maintained migration documentation across both the detailed and short “Upgrading” sections.

### DIFF
--- a/blog/releases/11.0.md
+++ b/blog/releases/11.0.md
@@ -9,7 +9,7 @@ pnpm 11 is here! This release tightens the security defaults introduced througho
 
 It also requires Node.js 22 or newer — pnpm itself is now pure ESM.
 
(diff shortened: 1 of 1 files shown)


zkochan:
The link is correct. It redirects to the right page. But if you change this now, it will break in the future and will lead to the next version's migration page.

zkochan:
ok, the redirect doesn't work for some reason via the single page app routing

maurer2:
Thanks. Weirdly enough the redirect works after reloading the page or when copy and pasting the URL. 

Cheers

PR #795, opened 2026-05-12
chore: decouple Crowdin sync from Vercel build

zkochan:
## Summary
- Commit current Crowdin export as `i18n/` (was `.gitignore`d) so Vercel builds from a known-good snapshot instead of pulling live from Crowdin on every deploy
- Pin `buildCommand` to `pnpm build` in `vercel.json`, overriding the project-level command that runs `crowdin:sync`
- Add `sync-translations.yml`: runs weekly (Mon 02:00 UTC) and on demand, runs `crowdin:sync`, builds with all locales as a smoke test, opens a PR with the result

## Why
Vercel's build command was `npm run crowdin:sync && npm run build`, which pushed sources to Crowdin with `--auto-update` and pulled a freshly-built translation ZIP on every deploy. That coupled prod deploys to the live state of a third-party system: a single malformed translation landing on Crowdin between two deploys could take down all future deploys, as happened recently with the zh-CN translation of ``--find-by &lt;finder_name\>`` in `_findBy.mdx` (where `&lt;` was replaced with a literal `<`, breaking Crowdin's renderer).

With this change, broken translations submitted on Crowdin can no longer break prod — at worst they fail the weekly sync PR's build, where they're caught before merge. `scripts/build-with-fallback.mjs` continues to handle individual broken translation files gracefully, falling them back to English.

## Required before merging
Add these secrets to the repo so the new workflow can run:
- `CROWDIN_PERSONAL_TOKEN` (already exists; used by `translations.yml`)
- `CROWDIN_PROJECT_ID` (new — value is `302994`)

Also: clear or update Vercel's project-level "Build Command" setting in the dashboard so it doesn't override what's now pinned in `vercel.json`.

## Test plan
- [ ] CI: PR's `Build Test` workflow passes (builds English only)
- [ ] Locally verify `pnpm build` succeeds with committed `i18n/` (confirmed: 4 pre-existing broken zh translations get auto-removed by `build-with-fallback.mjs`, otherwise green)
- [ ] After merging: trigger `sync translations` workflow manually and confirm it opens a PR
- [ ] Confirm Vercel deploy succeeds without running `crowdin:sync`


fucky4152-a11y (commented):
- - view_ltr.xlsx

PR #801, opened 2026-05-19
Add nx version disallow to example, intro "unreviewed" concept

karlhorky:
In the [Settings docs `allowBuilds` section](https://pnpm.io/settings#allowbuilds):

- [x] Add an example of disallowing specific `nx` versions
  - [x] Add comment describing what happens with other `nx` versions
- [x] Introduce "unreviewed" concept for versions that are NOT in the `allowBuilds` config, because this is not explicitly called out in the docs and is confusing to LLMs (and probably humans)


zkochan (approved):

karlhorky:
Thanks for the review and merge!

PR #802, opened 2026-05-19
chore: update cr new logo

santoshyadavdev:


zkochan:
i think you mixed up the one for dark theme and the one for light theme

santoshyadavdev:
> i think you mixed up the one for dark theme and the one for light theme

haha I also realized it, let me change it

zkochan (approved):

PR #828, opened 2026-06-20
feat(benchmarks): add Bun to benchmark generation

qeeqez:
# Adds Bun benchmarks to the comparison

Currently we are missing Bun from comparison.
This simply adds bun to it.

## Benchmark Results
- PNPM 12 is faster in firstInstall and repeatInstall
- Bun is faster in withWarmCacheAndLockfile, withWarmCache and else
So we could see the room of improvement for PNPM Rust port

image

## CI
After merging to main, "benchmark.yml" CI script will run benchmark suite and will regenerate bench results.
p.s. bench results are not included in PR because of it
p.p.s. current CI script already includes Bun setup, might be later switched to oven-sh/setup-bun@v2 action but lets do it in separate PR.

## Summary by CodeRabbit

* **New Features**
  * Bun is now included in the benchmark suite, with end-to-end measurement, installation scenarios, and reporting in charts and generated SVGs.

* **Bug Fixes**
  * Lockfile cleaning during benchmark runs now correctly handles Bun’s lockfile format.

* **Chores**
  * Benchmark documentation updated to list Bun alongside other package managers.
  * Removed a legacy npm benchmark results file.

### DIFF
--- a/benchmarks/benchmarkFixture.js
+++ b/benchmarks/benchmarkFixture.js
@@ -17,6 +17,7 @@ const lockfileNameByPM = {
   npm: 'package-lock.json',
   pnpm: 'pnpm-lock.yaml',
   pacquet: 'pnpm-lock.yaml',
+  bun: 'bun.lock',
   yarn: 'yarn.lock'
 }
 
@@ -255,4 +256,3 @@ function spawnSyncOrThrow (cmd, opts) {
   }
   return result;
 }
-
--- a/benchmarks/commandsMap.js
+++ b/benchmarks/commandsMap.js
@@ -45,6 +45,21 @@ export default {
       '--ignore-scripts',
     ]
   },
+  bun: {
+    scenario: 'bun',
+    legend: 'Bun',
+    color: '#f47216',
+    name: 'bun',
+    args: [
+      'install',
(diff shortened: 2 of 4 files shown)


zkochan:
You can see here all package managers comparisons if that is what you are after: https://benchmarks.vlt.sh/#/package-managers/average

It is now broken for pacquet.

I am not sure we need to mention bun on our website and provide free marketing to it. pnpm is currently more widely used than bun. We can have internal benchmarks if needed.

qeeqez:
@zkochan makes sense. And yeah, I see those broken results for parquet.

My point was to at least have unpublished test result (alotta-files.yml), so we could compare and make sure pnpm is going #1 performance wise on all fronts. 

You could close this PR, if you think so, no problem

---
I will try to check pnpm repo on what could be improved for:
- withWarmCacheAndLockfile
- withWarmCache
- withLockfile
- withWarmModulesAndLockfile
- updatedDependencies 

image

zkochan:
We have bencher.dev configured to make sure the performance of pnpm does not degrade. Also, i don't expect bun to become any faster. So it is enough to beat bun where we can then just keep that performance. I used the vlt benchmarks for that.

PR #841, opened 2026-07-18
docs: update pnpr docs for teams, staged publishing, and team endpoints

zkochan:
Reconciles the pnpr docs with the server changes that landed since they were last updated (2026-07-04).

## Config

- **`groups:` → registry-scoped `teams:`** ([#12790](https://github.com/pnpm/pnpm/pull/12790)). Teams are declared on the registry that uses them and referenced as `team:<name>`; a YAML anchor covers sharing one roster. The top-level `groups:` block no longer exists.
- **Access-token grammar** is now documented as a table plus its validation rules — notably that a bare token is always a username and never a team, so `access: platform` grants the *user* `platform` even when a team of that name exists.
- Added the `teams` key to the hosted and upstream registry tables.

## Endpoints

- **Staged publishing** ([#12922](https://github.com/pnpm/pnpm/pull/12922)) — the server half of [`pnpm stage`](https://pnpm.io/cli/stage), which was already documented on the CLI side. The two pages are now linked.
- **Team endpoints** ([#12789](https://github.com/pnpm/pnpm/pull/12789)) — read-only views over the config-declared teams; every mutation is a 403.
- **Conditional writes on S3** ([#12832](https://github.com/pnpm/pnpm/pull/12832)) — multi-replica safety and the 409 a conflicting writer sees.

## Drive-by fix

The upstream circuit-breaker keys are `maxFails` and `failTimeout`, not `max_fails` and `fail_timeout`. `UpstreamFile` is `rename_all = "camelCase"` with `deny_unknown_fields`, so the documented spellings were a hard startup error, not a silently-ignored key:

```
Invalid config: unknown field `max_fails`, expected one of url, public, auth,
headers, maxage, timeout, maxFails, failTimeout, cache, access, teams, packages
```

This predates the changes above and is unrelated to them.

## Verification

Checked against a real `pnpr` binary rather than only by reading the source:

- Every documented grammar rule exercised, including that a bare token matching a team name resolves as a username.
- Full staged-publish round trip: stage → package 404 while held → list → tarball → approve → package 200 → record dropped.
- Team endpoints hit live for their exact JSON shapes and the 403 on mutation.
- Every YAML example in the diff parses and boots.
- English Docusaurus build passes with no broken links or anchors from any pnpr page.

## Summary by CodeRabbit

* **Documentation**
  * Clarified registry package authorization rules, token grammar/validation behavior, and registry-scoped team configuration (including access defaults).
  * Expanded registry endpoint documentation, including staged publishing workflows, approval-time conflict handling, and visibility/status code semantics.
  * Added/defined read-only npm team endpoint behavior with privacy-preserving denial responses and empty results when no teams exist.
  * Documented hosted storage details for staged packages and replica-safe concurrent publishing on S3 backends.
  * Updated upstream registry configuration fields, including circuit-breaker setting name casing.

### DIFF
--- a/pnpr-docs/configuration.md
+++ b/pnpr-docs/configuration.md
@@ -118,11 +118,35 @@ per specificity tier can match a name, so the winning rule is unique and
 reordering the map (by a formatter, or a YAML round-trip) can never change
 which rule applies. A duplicate key is a config error.
 
-Each value can set `access`, `publish`, and `unpublish`. Common values are
-`$all` (anyone), `$authenticated` (logged-in users), and `$anonymous`; a list
-can also name users and [groups](#groups). When a field is omitted, `access`
-falls back to the registry-level `access:` default, `publish` defaults to
-`$authenticated`, and `unpublish` defaults to nobody.
+Each value can set `access`, `publish`, and `unpublish`. When a field is
+omitted, `access` falls back to the registry-level `access:` default, `publish`
+defaults to `$authenticated`, and `unpublish` defaults to nobody.
+
+Every one of those fields takes either a single access token or a YAML list of
+them, and the grammar is deliberately small:
+
+| Token | Admits |
+| --- | --- |
+| `$all` | Anyone, authenticated or not. |
+| `$authenticated` | Any logged-in user. |
+| `$anonymous` | Callers with no credentials. |
+| `team:<name>` | The members of a [team](#teams) declared on this registry. |
+| `alice` | Exactly that username. Any bare token is a username. |
+
+Tokens are validated at config load and every near-miss is rejected loudly,
+rather than being read as a username that happens to admit nobody:
+
+- `$` is reserved for the three built-ins — any other `$token` is an error.
+- `all`, `authenticated`, `anonymous`, and `@`-prefixed spellings like `@all`
+  are errors suggesting the `$` form.
+- `team:` is the only typed prefix; any other `<kind>:` prefix is an error.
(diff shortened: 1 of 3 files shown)


zkochan:
Thanks — both findings were valid and are fixed in f75411a1.

**Retry wording.** Correct, and my own `storage.md` text contradicted it. `update_hosted_packument_with_retry` has exactly two callers: `update_dist_tag` on the request path, and journal roll-forward during recovery. Publish and partial unpublish use the conditional write without a retry, so a lost race is a 409. Reworded to distinguish the two.

**Staging vs. approval, and 2FA.** Also correct, and stronger than a nitpick — the old wording implied a capability pnpr doesn't have. pnpr has no OTP support at all; the only `2FA` reference in the crate is a comment noting the profile endpoint doesn't track it. `pnpm stage approve --otp` exists for registries that raise a challenge, which pnpr never does. Split the paragraph into what each phase checks and dropped the proof-of-presence claim.

Verified the new claims against a running `pnpr` rather than only by reading:

```
1) stage v1.0.0:   {"ok":true,"stageId":"8d20b8f7-…"}
2) approve:        {"ok":true} [201]
3) re-stage the same, now-published version:
                   {"ok":true,"stageId":"ccdbd7c5-…"}  201   <- no conflict check at staging
4) approve it:     Cannot publish over the previously published version
                   @mycompany/demo@1.0.0  [409]              <- conflict caught at approval
5) record survives the 409 and stays listable
```

Left alone: the `packument` spelling LanguageTool flagged is the correct domain term and already used throughout these docs.

PR #845, opened 2026-07-27
docs: copy pnpm-workspace.yaml before pnpm install in Docker

Booyaka101:
### What

The v11 Dockerfile example in `docs/docker.md` copies only `package.json` and `pnpm-lock.yaml` before `pnpm install`:

```dockerfile
COPY package.json pnpm-lock.yaml ./
RUN pnpm install --frozen-lockfile
```

Since v11, `pnpm-workspace.yaml` is the per-project configuration file — [`docs/settings.md`](https://pnpm.io/settings) says settings moved out of `package.json#pnpm` and out of `.npmrc` into it. So this install runs with `nodeLinker`, `hoistPattern`, `overrides`, `allowBuilds` and everything else **not applied**.

It does not fail. The image builds, the container starts, and the dependency tree is quietly not the one the project configured — which makes it easy to miss after upgrading, especially since the `pnpm-v10-to-v11` codemod *creates* `pnpm-workspace.yaml` for projects that previously had none.

### Changes

- `docs/docker.md` — add `pnpm-workspace.yaml` to the `COPY` step, with a note on why and a caveat for projects that do not have one.
- `docs/migration.md` — add it as a manual follow-up, next to the existing `npm_config_*` item that already points at Docker images.

Only the one example that copies files selectively is affected; the `COPY . .` and `pnpm fetch` recipes elsewhere on the page are already fine.

### Notes

Happy to switch the example to a `pnpm-workspace.yaml*` glob instead if you would rather it stay copy-pasteable for projects without the file — I went with the plain form to match the surrounding style, and covered that case in the note.

## Summary by CodeRabbit

- **Documentation**
  - Updated Docker setup guidance to copy `pnpm-workspace.yaml` into the image before running `pnpm install --frozen-lockfile`.
  - Clarified that without copying the workspace file first, pnpm may silently ignore workspace settings (from `package.json#pnpm`/`.npmrc`) during installation.
  - Added a Docker migration note for upgrades to version 11.
  - Noted that the workspace file can be omitted if the project doesn’t use `pnpm-workspace.yaml`.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -60,11 +60,20 @@ Or let pnpm install Node.js automatically from [`devEngines.runtime`](./package_
 ```dockerfile
 FROM ghcr.io/pnpm/pnpm:11
 WORKDIR /app
-COPY package.json pnpm-lock.yaml ./
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
 RUN pnpm install --frozen-lockfile
 COPY . .
 CMD ["pnpm", "start"]
 ```
+
+:::note
+
+`pnpm-workspace.yaml` is the [per-project configuration file](./settings.md), so it
+must be copied in before `pnpm install` runs. Without it the install still succeeds,
+but silently ignores every setting it contains — `nodeLinker`, `hoistPattern`,
+`overrides`, `allowBuilds` and so on. Omit it if your project does not have one.
+
+:::
 
 ### When to use this image
 
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -68,10 +68,12 @@ CMD ["pnpm", "start"]
 
 :::note
 
-`pnpm-workspace.yaml` is the [per-project configuration file](./settings.md), so it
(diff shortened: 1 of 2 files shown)


Booyaka101:
Thanks — the contradiction was real and mine. Pushed 9a7d749.

**Fixed:** `migration.md` said Docker images "must" copy `pnpm-workspace.yaml` unconditionally, while the `docker.md` note said to omit it when absent. Both now phrase it conditionally ("if your project has one"). I also spelled out that in a workspace the file declares `packages`, so omitting it there resolves the wrong project set rather than merely dropping settings — which is the case the previous wording glossed over.

**Not changed — `.npmrc` / private registries.** Correct that this example copies no `.npmrc` and that v11 still reads auth and registry settings from it, but that is pre-existing on this page and independent of the workspace file. Secret-safe `.npmrc` injection in Docker is a genuinely separate topic (build secrets, not a `COPY`), and folding it in would take a two-line fix well past its scope. Happy to open a follow-up PR for it if you'd like that documented.

PR #850, opened 2026-07-31
docs: document pnpm 12 and label the docs version "11 & 12"

zkochan:
pnpm 12 is a Rust rewrite with no intentional breaking changes vs v11, so it doesn't need its own docs version — the 11.x docs already describe it. Two changes follow from that:

1. **Relabel the 11.x docs version to "11 & 12"** so the version dropdown says what the docs actually cover.
2. **Document the beta's installation** in a dedicated section of the existing installation page, rather than a separate `installation-v12` page. v11 stays the default path throughout.

## The relabel

```ts
versions: {
  [lastDocsVersion]: { label: '11 & 12' },
},
```

Only the *displayed* label changes — the version is still named `11.x` on disk, in URLs, in the `versioned_sidebars` filename and in the Crowdin paths. So there's no snapshot to commit, no `.gitignore`/`copy-docs` rename, no second sidebar file, and no translation churn. The dropdown and the "for up-to-date documentation" banner on the 10.x pages both pick it up.

Cutting a real `12.x` version later is still one `git add -f versioned_docs/version-11.x` away if v12 ever diverges.

## Install paths, in order

1. **`pnpm self-update next-12`** — leads the section. Anyone already on pnpm 11.10.0+ needs nothing else, and it's already documented at `docs/cli/self-update.md:42`.
2. **npm**, for people without pnpm.
3. **The standalone script**, for people without Node.js.

## Verified against pnpm 12.0.0-beta.2

Every claim below was checked against the published artifacts, not inferred:

- **`pnpm self-update next-12` works** — ran it against an isolated `PNPM_HOME` from pnpm 11.18.0; it linked the native binary and the result runs.
- **npm needs `--allow-scripts=pnpm`.** The published package is a wrapper whose `preinstall` script swaps in the native binary from `@pnpm/exe.<platform>`. npm 12 blocks install scripts by default, so a plain `npm install -g pnpm@next-12` silently leaves the placeholder and `pnpm` fails with a shell syntax error. The flag is accepted (and harmless) on npm 11 too, so one command covers both.
- **The standalone script needs an exact `PNPM_VERSION`.** `install.sh` interpolates it straight into the GitHub release URL, so a dist-tag like `next-12` would 404. (`install.ps1` does resolve dist-tags, but documenting the exact version keeps both platforms identical.)
- **Corepack cannot install v12.** It looks for `bin/pnpm.mjs`, which the native package doesn't ship — it fails with `MODULE_NOT_FOUND`. Documented as a known limitation.
- **Intel macOS works again on v12** (`pnpm-darwin-x64.tar.gz` is in the release, and get.pnpm.io's abort is already `major_version -eq 11`), so the existing `darwin-x64` warnings are now scoped to v11.
- **`pnpm` and `@pnpm/exe` are identical on v12** — same bins, same optional deps — so the v11-only distinction is labelled as such.

## Also

- Compatibility table gains a `pnpm 12` column (`engines: node >=18`), with a note that Node.js is only needed to *install* from npm, not to run pnpm.
- `/12.x/*` → `/*` redirect for guessed URLs, deliberately non-permanent so it can be dropped if `12.x` ever becomes a real docs version.

`LOCALE_CI=en docusaurus build` passes; the new anchors resolve and `/installation` is absent from the (pre-existing) broken-anchor list.

## Open questions

- **Translated locales may keep showing "11.x"** until Crowdin picks up the new label. Version labels are translatable (`i18n/<locale>/docusaurus-plugin-content-docs/version-11.x.json`), and existing entries hold the old string. I couldn't verify this locally — there's no `i18n/` checkout here — so it's worth a look after the next `crowdin:sync`.
- In my test, `pnpm self-update next-12` installed **12.0.0-beta.0**, not the `next-12` tag's current 12.0.0-beta.2 — the release-age cooldown resolving to the newest eligible version. Expected behavior, and `docs/cli/self-update.md` covers `minimumReleaseAge`, but it may surprise beta testers. Say the word if you want a note about it on the page.
- `12.0.0-beta.2` is hardcoded in the standalone-script commands and will need bumping each beta, as was done for the v11 RCs in 99943f97.
- The **"Linux runtime requirements"** note (glibc 2.27+, `libatomic.so.1`) is left untouched — it was written for the v11 Node SEA build and I couldn't confirm what the Rust binary actually needs. Worth a follow-up if the requirements changed.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -7,6 +7,12 @@ title: Installation
 
 If you don't use the standalone script or `@pnpm/exe` to install pnpm, then you need to have Node.js (at least v22) to be installed on your system.
 
+:::info
+
+Looking for pnpm 12? It is currently in beta and installed differently from pnpm 11. See [Installing the pnpm 12 beta](#installing-the-pnpm-12-beta).
+
+:::
+
 ## Using a standalone script
 
 You may install pnpm even if you don't have Node.js installed, using the following scripts.
@@ -38,7 +44,9 @@ Add-MpPreference -ExclusionPath $(pnpm store path)
 
 :::warning Not supported on Intel macOS
 
-The standalone script does not run on Intel Macs (`darwin-x64`). Use [npm](#using-npm), [Corepack](#using-corepack), or [Homebrew](#using-homebrew) instead. See [#11423](https://github.com/pnpm/pnpm/issues/11423) for context.
+On pnpm 11, the standalone script does not run on Intel Macs (`darwin-x64`). Use [npm](#using-npm), [Corepack](#using-corepack), or [Homebrew](#using-homebrew) instead. See [#11423](https://github.com/pnpm/pnpm/issues/11423) for context.
+
+pnpm 12 ships an Intel macOS build again, so this limitation doesn't apply to it.
 
 :::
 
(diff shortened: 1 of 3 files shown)


zkochan:
Both review comments addressed in b33d9d2b.

**Intel macOS warning title** — agreed, the title contradicted its own body. Now reads "Not supported on Intel macOS in pnpm 11".

**npm version requirement** — partly. I verified the documented command against npm 10, 11 and 12 in isolated prefixes: it installs 12.0.0-beta.2 and produces a working native binary on all three. On npm older than 11.16 the flag is simply ignored (install scripts run by default there), so no version gate or alternative command is needed. But the *reason* for the flag was vague, so it now names the version that introduced the behavior and says explicitly that older npm ignores it.

I did not add Node.js 18 routing advice: `engines` on the v12 wrapper is `>=18`, and the Node.js version doesn't affect whether the flag is needed — only the npm version does.

PR #851, opened 2026-07-31
docs: split the settings page so Algolia can index it again

zkochan:
## Why

`/settings` is not in the Algolia index at all. Querying the live index for `docs-default-11.x` returns **0 records** whose URL is `/settings`, while the 10.x copy of the same page has 68+. That is why searching for `packageImportMethod`, `dedupePeers`, `nodeDownloadMirrors`, `requiredScripts` or `optimisticRepeatInstall` returns nothing — the only hits for names like `hoistPattern` come from other pages that happen to mention them.

The cause looks like page size. The rendered page had grown to **237 KB**:

| page | rendered | indexed |
|---|---|---|
| `/settings` (11.x) | 237 KB | ❌ |
| `/10.x/settings` | 186 KB | ✅ |
| `/pnpmfile` (next biggest on the site) | 161 KB | ✅ |

So the crawler's document ceiling sits between 187 KB and 237 KB and we crossed it. The exact number is in the crawler dashboard's error log, but the fix is the same either way — and it should land well under 187 KB, not just barely under.

## The split

Nine pages under `/settings/*`, along the existing `##` groups. No content was rewritten and no heading was renamed, so **all 145 anchors are preserved** — verified against the ids in the previously built page.

| page | groups | rendered |
|---|---|---|
| `/settings/dependency-resolution` | Dependency resolution | 83 KB |
| `/settings/node-modules` | node_modules + hoisting | 53 KB |
| `/settings/other` | Other | 51 KB |
| `/settings/build` | Build | 46 KB |
| `/settings/peer-dependencies` | Peer dependencies | 45 KB |
| `/settings/cli` | CLI + Node.js | 40 KB |
| `/settings/store` | Store + lockfile | 40 KB |
| `/settings/versioning` | Versioning | 38 KB |
| `/settings/network` | Network + requests | 35 KB |
| `/settings` (index) | — | 60 KB |

Pages holding two groups keep both group headings, so the index links to a group rather than a page and the old section anchors (`#lockfile-settings`) still resolve.

## Old deep links

`pnpm.io/settings#hoistpattern` is printed by pnpm itself and linked from issues, blog posts and third-party articles. A URL fragment never reaches the server, so `vercel.json` can't redirect these — it has to happen in the browser.

`src/components/SettingsAnchorRedirect.js` (mounted on the index page only) takes the requested anchor, finds the index entry that links to it, and forwards to the page that documents it now. The rendered index *is* the map, so it cannot go stale: a setting missing from it is a setting missing from the docs. Section anchors still exist as headings on the index and are left to scroll natively.

Checked against the build output: every one of the 145 original anchors resolves, and every setting has an index link, so no old deep link is left without a target.

## `pnpm-workspace.yaml` page merged in

The settings page is named after the file and opens by explaining it, so a separate page describing the same file was a step rather than information. Its two sections, `packages` and `packageConfigs`, now sit at the top of `/settings` ahead of the index, and `/pnpm-workspace_yaml` 301s to `/settings` (fragments are preserved by the browser across the redirect, so `#packageconfigs` still lands correctly).

The index also claimed to list every setting while eleven workspace settings live on the workspaces page and a few others are documented with the command or file they belong to. Those are listed at the end now, which additionally gives the anchor redirect a target for old links like `/settings#linkworkspacepackages` — broken before this PR, working after it.

## Also

- 69 in-repo doc links + 36 blog links repointed at the new pages. Anchors that were already broken and had a real home are fixed too: `#linkWorkspacePackages` / `#includeWorkspaceRoot` → `/workspaces`, `#patcheddependencies` → `/cli/patch`, `#managepackagemanagerversions` → `#pmonfail`, plus a few case-mismatched ones (`#nodeLinker`, `#storeDir`).
- Sidebar entry becomes a category linked to the index page, in both `sidebars.json` and `versioned_sidebars/version-11.x-sidebars.json`.
- `LOCALE_CI=en docusaurus build` passes. No settings-related broken link or anchor remains; what's left is pre-existing and unrelated (`/cli/pn` in the 11.0 post, `/npmrc#side-effects-cache`, `#onlybuiltdependencies` in a 10.19-era post) plus 10.x, which is untouched.

## After merging

- Trigger a crawl (or wait for the scheduled one) and confirm records appear, e.g. `curl` the index for `packageImportMethod` filtered to `docusaurus_tag:docs-default-11.x`.
- The new files reach Crowdin as new sources. Content is unchanged, so a **pre-translate from TM** run should restore the existing translations; worth doing right after the upload so the translated pages don't fall back to English.

## Summary by CodeRabbit

* **Documentation**
  * Reorganized configuration guidance into dedicated pages covering build, CLI, dependency resolution, networking, Node modules, peer dependencies, storage, versioning, and other settings.
  * Updated release notes, guides, and CLI references to link to the new settings locations.
  * Consolidated workspace configuration documentation into the settings reference.
* **Navigation**
  * Updated configuration sidebars with grouped settings pages.
  * Added redirects and anchor handling to preserve access to existing settings links.
  * Added a permanent redirect for the former workspace configuration page.


zkochan:
Thanks — went through the review comments. Fixed in cebc5673:

- **`minimumReleaseAgeStrict` anchor** (`docs/cli/pnx.md`) — it pointed at `#minimumreleaseage`, so the link opened the wrong section. Predates this PR; fixed since the line was already being touched.
- **`ini` fence on a `pnpm-workspace.yaml` snippet** — now `yaml`. Also raised "even if someone is using Node.js v16" to v22 in the same section, since pnpm 11 can't run on v16.
- **Conflicting-peer example** — the symlinks pointed at `webpack@1.0.0` and `webpack@1.0.0_esbuild@1.0.0`, neither of which appears in the `.pnpm` tree drawn two lines above. They now point at the suffixed directories that are listed.
- **`hashchange` in `SettingsAnchorRedirect`** — good catch: the page stays mounted when only the fragment changes, so the effect never re-ran. It now redirects on mount *and* on `hashchange`, with cleanup.

Three I'm not acting on:

- **Missing `versioned_docs/version-11.x/settings/*.md`** — false positive. That directory is gitignored and regenerated on every build by `copy-docs` (`rm -rf versioned_docs/version-11.x && cp -r docs versioned_docs/version-11.x`), which `pnpm build` runs before Docusaurus. The build passes.
- **Heading levels skipping H2** — deliberate. Five settings on `/settings/other` come from `_*.mdx` partials that `cli/run.md`, `catalogs.md` and others include under their own H2 sections, where H3 is the correct level. Promoting the settings to H2 would either break those pages or leave `other.md` with sibling settings at mixed levels, which is worse than the skip. Pages that hold two groups do have H2 group headings.
- **Dependency visibility vs hoisting** (`extendNodePath` → "Why this is needed") — the "a package can only access its own declared dependencies" framing is used across the docs and this PR doesn't change that text. Worth a separate look if it should be qualified for `hoistPattern`, but not something to settle inside a reorganization PR.

Unrelated: the failing Vercel check isn't from this branch — every open PR (#848, #846, #845, #839) shows the same preview failure while `main` deploys fine.

PR #853, opened 2026-08-04
docs: document pnpm 11.20

zkochan:
Documents the changes released in pnpm v11.20.0.

## Docs

- **`namedRegistries`** ([settings/dependency-resolution](https://pnpm.io/settings/dependency-resolution#namedregistries)) gains three subsections:
  - *Built-in aliases* — `gh:` plus the new `npmjs:`, how user entries override them, why `npm:` cannot pin to the public registry, and the tarball-URL-prefix consequence for orgs that mirror npmjs.
  - *Reserved alias names* — `ERR_PNPM_RESERVED_NAMED_REGISTRY_NAME` and the reserved prefix list.
  - *Named registries in the lockfile* — registry-qualified keys (`foo@work:1.0.0`), the package-substitution risk they fix, the unchanged lockfile format version, a caution covering the migration (commit the diff, get everyone onto v11.20+ first, no opt-out), and `ERR_PNPM_MISSING_NAMED_REGISTRY`.
- **Network settings** — empty / `false` / `null` proxy values now read as unset instead of failing with `ERR_PNPM_INVALID_PROXY`; CLI values stay verbatim; the legacy `.npmrc` `proxy` fallback (#13533).
- **`pnpm licenses`** — packages are reported per registry; alias shown in the table and as `registryName` with `--json`.
- **`pnpm sbom`** — purl `repository_url` qualifier, including what is stripped from the URL.
- **Supported package sources** — new "Named registries" trusted-source section.
- **Mitigating supply chain attacks** — new section on pinning dependencies to the registry they come from.

## Blog

Adds `blog/releases/11.20.md`, covering the named-registry security fix, the `npmjs:` alias, and the patch changes (proxy handling, `pnpm rebuild` path-traversal rejection, the `@pnpm/exe` env-lockfile fix, tarball-URL match ordering, and the resolution speedups).

Verified against the v11.20.0 release commit and the pnpm source. `pnpm build` passes; the only broken link it reports is the pre-existing `/cli/pn` in the 11.0 release post.

## Summary by CodeRabbit

* **New Features**
  * Added registry-qualified lockfile entries, built-in `gh:` and `npmjs:` aliases, and improved deterministic tarball matching.
  * Improved license and SBOM reporting for packages from named registries.
  * Enhanced proxy handling, dependency-resolution performance, and environment lockfile support for pnpm 12+.
  * Added clearer validation for missing or reserved registry aliases.

* **Security**
  * Strengthened `pnpm rebuild` protections and guidance for pinning dependencies to their source registry.

* **Documentation**
  * Expanded guidance for named registries, lockfile migrations, package sources, proxy settings, licenses, and SBOM output.

### DIFF
--- a/blog/releases/11.20.md
+++ b/blog/releases/11.20.md
@@ -13,13 +13,13 @@ pnpm 11.20 fixes a package-substitution risk in projects that install from more 
 
 ### Named registries are recorded in the lockfile
 
-**Security fix.** This affects projects using [`namedRegistries`](/settings/dependency-resolution#namedregistries) on pnpm 11.1.0–11.19.x, and is semi-breaking for them — see "If you use named registries" below.
+**Security fix.** This affects projects that install through a [named registry](/settings/dependency-resolution#namedregistries) alias on pnpm 11.1.0–11.19.x — an alias you configured under `namedRegistries`, or the built-in `gh:`, which needs no configuration. It is semi-breaking for them; see "If you use named registries" below.
 
 The lockfile recorded no marker for which registry a package came from. Packages were keyed by `name@version` alone, so when two registries served the same name and version, both collapsed onto a single `packages:` entry and whichever resolved first decided the tarball that every consumer got. A package you expect from your private registry could be installed from a different registry that publishes the same name and version, with nothing in the lockfile to reveal it.
 
 Packages resolved from a named registry are now recorded under registry-qualified keys — `<name>@<registryName>:<version>`, e.g. `foo@work:1.0.0` — so each registry gets its own entry and the lockfile pins which one a dependency came from.
 
-The lockfile format version is unchanged. Registry-qualified keys appear only for packages resolved from a named registry, so a project that does not use `namedRegistries` sees no difference, and older pnpm versions keep reading the file.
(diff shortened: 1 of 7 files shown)


zkochan:
Addressed in 16b43d04:

- **`docs/package-sources.md` (fenced block)** — tagged the example block as `sh`.
- **`docs/package-sources.md` (config location)** — the alias mapping note now mentions the global `config.yaml` alongside `pnpm-workspace.yaml`, with the v11.11.0 version context.
- **Built-in aliases and the lockfile migration scope** — good catch, this was imprecise. `mergeNamedRegistries` merges the built-ins into the same map the named-registry resolver reads, so a package installed via `gh:` gets a registry-qualified key (`foo@gh:1.0.0`) with no `namedRegistries` entry configured. Rewrote the scope in both `docs/settings/dependency-resolution.md` and the release post: qualified keys apply to any named-registry alias including the built-in `gh:` and `npmjs:`, and the "no difference" case is now a project that installs nothing through an alias. `gh:` has been built in since v11.1.0, so it is in scope for the re-keying; `npmjs:` is new in v11.20.0 and has no pre-existing lockfile entries to migrate.

Skipped the LanguageTool "GitHub" capitalization hint — `github` there is a literal reserved dependency specifier prefix, not the product name.

PR #856, opened 2026-08-04
docs: document `frozenLockfile` setting

### LINKED ISSUE #13643: Add workspace setting to force `--frozen-lockfile`
### Contribution

- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/pnpm/pnpm/blob/main/CONTRIBUTING.md))

### Describe the user story

As a developer, I want to ensure that my team is always safe when installing dependencies and does not risk pulling in anything other than what the lockfile specifies, even if they accidentally modify the package.json file, etc.

### Describe the solution you'd like

I'd like to see a config option that can be set for the workspace that acts the same as the existing `--frozen-lockfile` flag, but without needing to remember to use the flag.

Yarn has a similar flag already: https://yarnpkg.com/configuration/yarnrc#enableImmutableInstalls

### Describe the drawbacks of your solution

Updating the package.json file and using the install command to update the lockfile will not work, dependency changes will need to be done via the dedicated add/remove commands.

### Describe alternatives you've considered

_No response_

MattIPv4:
Closes https://github.com/pnpm/pnpm/issues/13643

## Summary by CodeRabbit

* **Documentation**
  * Added `frozenLockfile` to the lockfile settings overview.
  * Documented default behavior across CI and non-CI environments.
  * Clarified when installation fails due to a missing or out-of-date lockfile.
  * Explained how the setting relates to CI behavior.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -184,6 +184,7 @@ Every setting is listed below, grouped by topic. Follow a setting to read its do
 [Full reference →](./settings/store.md#lockfile-settings)
 
 * [lockfile](./settings/store.md#lockfile)
+* [frozenLockfile](./settings/store.md#frozenlockfile)
 * [preferFrozenLockfile](./settings/store.md#preferfrozenlockfile)
 * [lockfileIncludeTarballUrl](./settings/store.md#lockfileincludetarballurl)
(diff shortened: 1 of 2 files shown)


MattIPv4:
👋 @zkochan checking in on if this can be landed?

PR #864, opened 2026-08-10
docs: document globalShims and project-aware global bins

zkochan:
Documents the `globalShims` behaviour released in pnpm **v12.0.0-rc.2** (pnpm/pnpm#13731).

## Changes

**`docs/settings/other.md`** — new `globalShims` reference entry after `globalBinDir`:

- Default `{ node: auto, deno: auto, bun: auto }`; type Boolean or Object
- Keyed by the *providing package* name, so an entry for `typescript` covers its `tsc` bin
- Policy table: `auto` (or `true`), `prompt`, `always`, `false`
- Key-wise merging across layers vs. the scalar shorthands that replace the whole map
- A warning that the dispatcher reads this **only** from locations a project cannot write to — the global config file, the pnpm home's own `pnpm-workspace.yaml`, and `PNPM_CONFIG_GLOBAL_SHIMS` — never a project's `pnpm-workspace.yaml`
- `PNPM_SHIM_BYPASS=1` for a one-off bypass

**`docs/global-packages.md`** — new "Project-aware global bins" section covering the user-facing behaviour: the `cd` into a pinned project → `node --version` example, how a version is chosen (runtimes resolve from `devEngines.runtime` / `engines.runtime` only and never from `node_modules/.bin`; other packages resolve from `node_modules/.bin`), and the trust model — signature-verified stable Node runs promptless, everything else prompts once per project *and* binary fingerprint, non-interactive/CI falls back to the global version, and same-package validation defeats lookalikes.

**`docs/package_json.md`** — one line under `devEngines.runtime` noting that a bare `node` now follows the pin, not just scripts.

## Notes

- **One detail differs from the feature commit message.** It says the mode "takes effect without relinking", which is only half true: the dispatcher re-reads the setting on every dispatch, so *disabling* or changing a policy is immediate, but `global.rs` partitions bins into direct vs. context-aware shims at **install** time, so newly *enabling* a package needs `pnpm add -g <pkg>` to rewrite its shim. Both halves are documented.
- **Version labelling.** `docs/` serves the version labelled "11 & 12", so the feature is marked `Added in: v12.0.0-rc.2` with an explicit "not available in v11" note, following the existing `Added in:` convention.

## Verification

`pnpm build` passes. The new anchors and all three cross-links resolve in the built HTML. The only broken links reported are the pre-existing `/cli/pn` reference in the 11.0 blog post across locales, untouched here.

## Summary by CodeRabbit

* **Documentation**
  * Documented project-aware global shims for Node.js, Deno, and Bun commands.
  * Added guidance for runtime version resolution, trust prompts, package policies, CI behavior, and bypass options.
  * Documented the `globalShims` setting, including configuration formats, defaults, trusted sources, and reinstall requirements.
  * Clarified that bare runtime commands use versions pinned through `devEngines.runtime`.
  * Added guidance for disabling or bypassing shims and noted v11 incompatibility.

### DIFF
--- a/docs/global-packages.md
+++ b/docs/global-packages.md
@@ -172,6 +172,8 @@ If the project provides nothing, or the lookup is declined, the globally install
 
 A project you `cd` into is not automatically allowed to run its own binaries in place of your global ones.
 
+This section describes the default `auto` policy. Under `auto`:
+
 * A **stable Node.js release** is verified against the Node.js release team's signatures before it runs, so it switches without asking. (On musl-based systems the matching builds are unsigned, so they fall under the prompt instead.)
 * **Everything else** — Deno, Bun, Node.js prereleases, and any ordinary package you enable — asks once, per project and per candidate:
 
@@ -185,7 +187,9 @@ Answers are remembered in a machine-local registry, keyed to the project directo
 
 In CI and any non-interactive session, the question cannot be asked, so the global version runs and nothing is recorded.
 
-There is one more guard that applies regardless of trust: the project's command must come from the **same package** as the global one. A project that ships a lookalike `tsc` from some other package does not match, and your global `tsc` runs.
+The other two [policies](./settings/other.md#globalshims) change which candidates reach that question. `prompt` sends every candidate through it, including a signature-verified stable Node.js release; answers are still remembered, so it asks once per project and candidate rather than on every run. `always` skips the question entirely and switches straight away — which is also what makes it usable in CI, where a prompt would otherwise fall back to the global version.
+
+There is one more guard that applies regardless of policy: the project's command must come from the **same package** as the global one. A project that ships a lookalike `tsc` from some other package does not match, and your global `tsc` runs.
 
 ### Which packages participate
 
--- a/docs/global-packages.md
+++ b/docs/global-packages.md
@@ -126,6 +126,94 @@ You can check the current global bin directory with:
 pnpm bin -g
 ```
 
+## Project-aware global bins
+
+Added in: v12.0.0-rc.2 (pnpm v12 only)
+
+Global commands can follow the project you are standing in. When you run a global command from inside a project that asks for a different version of the same tool, pnpm runs the version the project asks for.
+
+The most useful case is Node.js itself. Given a project that pins its runtime:
+
+```json title="package.json"
+{
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": "^22.0.0",
+      "onFail": "download"
+    }
+  }
+}
+```
+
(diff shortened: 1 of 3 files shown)


zkochan:
Thanks — addressed two of the three points in eb7504e8.

**Trust policies scoped (`docs/global-packages.md`)** — done. The walkthrough now says up front that it describes the default `auto` policy, and a new paragraph covers how `prompt` and `always` differ.

One correction to the suggested wording, though: `prompt` does **not** "ask every time". It falls through to `is_trusted()`, which reads the recorded decision first and only prompts when there is no record ([`shim_dispatch.rs:162`](https://github.com/pnpm/pnpm/blob/8a453e36cf/pnpm/crates/cli/src/shim_dispatch.rs#L162)). So it asks once per project and candidate, exactly like `auto` — the real difference is that it does not exempt signature-verified stable Node.js from the gate. Documenting it as prompting on every run would have been wrong, so both the prose and the policy table in `settings/other.md` now state the gate-vs-frequency distinction. I also noted that `always` is the policy that actually works in CI, since a prompt there falls back to the global version.

**Bypass mechanisms (`docs/package_json.md`)** — done. The `devEngines.runtime` note now mentions both `globalShims` and `PNPM_SHIM_BYPASS=1`.

**`engines.runtime` precedence** — declining this one; the docs are correct as written. The dispatcher reads *both* fields for the runtime pin, `devEngines.runtime` first and `engines.runtime` second:

```rust
for engines_field in ["devEngines", "engines"] {
    let Some(runtime) = manifest.get(engines_field).and_then(|field| field.get("runtime"))
```

— [`shim_dispatch.rs:430`](https://github.com/pnpm/pnpm/blob/8a453e36cf/pnpm/crates/cli/src/shim_dispatch.rs#L430). `engines.runtime` here is a `{ name, version }` entry read from the project's own manifest, not `engines.node` and not a dependency's requirement, so the fallback is real and the precedence as documented matches the implementation.

---

Unrelated to this PR: the **Vercel check failure is pre-existing infrastructure**, not this change. It fails during `pnpm install`, before any docs are processed:

```
Using pnpm@9.x based on project creation date
ERROR  packages field missing or empty
Error: Command "pnpm install" exited with 1
```

Vercel ignores the `packageManager: pnpm@11.8.0` field and picks pnpm 9 by project creation date; pnpm 9 then rejects a `pnpm-workspace.yaml` that has no `packages` field. Every Preview deployment has failed this way in 4–7s for at least the past 6 days, while Production deployments succeed — and this commit only adds lines to three `.md` files. Worth a separate fix (enable corepack on Vercel, or pin the install command).

`pnpm build` passes locally; the new anchors and cross-links resolve in the built HTML.

PR #939, opened 2026-09-18
docs: document `pnpm cache prune`

TrevorBurnham:
Documents the CLI changes in https://github.com/pnpm/pnpm/pull/15048.

## New page: `cli/cache-prune`

`pnpm cache prune` removes the registry metadata cache directories the running version can no longer read. v12.4.0 changed the directory name for every registry, and `list`, `view` and `delete` all scope their search to the configured registry's current name, so the directory left behind was unreachable. The page covers the output format, `--dry-run`, and the three things prune deliberately keeps: the descriptor-scoped roots under `v11/metadata-private`, names already in the current shape including the pile-up from registries on an ephemeral port, and the sha256 an over-long registry URL collapses to.

It carries a `:::warning` about the cache being shared machine-wide. The naming change shipped in v11.27.0 as well as v12.4.0, so a pnpm at v11.26 or earlier under the same `cacheDir` still reads what prune removes and pays one metadata refetch for it.

New command, so v12 only, matching the code. `versioned_sidebars/version-11.x-sidebars.json` is untouched.

## `cli/cache-list-registries`

The 12.x page already said the command prints full URLs, which was not true of the code before 15048; it printed the encoded directory name. Both copies now show sample output and name the version where the output became URLs. The v11 copy also picks up the v11.27.0 directory naming change, which its page never mentioned.

`pnpm cache list` still prints directory names, and both pages keep saying so.

## Versions

`Added in: v12.5.0`, and v11.27.1 for the list-registries fix, are the next releases the two changesets in 15048 would land in. Neither is out yet, so these want a check before merging.

`pnpm build` is clean. The one broken link it reports, `/cli/pn` from `blog/releases/11.0`, predates this branch.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation for the experimental `pnpm cache prune` command, including supported metadata directories, the `--dry-run` option, and retained cache data.
  * Clarified that `pnpm cache list-registries` prints one registry URL per line, with examples and historical output differences.
  * Added `cache prune` to the cache command listings and Manage cache navigation for current and versioned documentation.


zkochan:
Superseded by merged PR #940, which incorporated this documentation and aligned it with the shipped pnpm v12.5.0 code. Thank you!

